Issue #1, opened 2018-04-13
Create project skeleton


bzz:
Update project structure with the setup used for applicastion

 - [ ] FE: a `package.json` \w standard scripts that do `echo 1`
 - [x] BE: Empty test. Almost the same code structure as in CAT (use chi, but a separate dir for FE)
 - [x] Build structure, single binary
 - [x] TravisCI

bzz:
> moved this from backlog to In Progress in Applications 10 days ago

do you think I would be a good time to start submitting the work you has done on this as PRs?  Do not need to be complete, etc

dpordomingo:
Yes, I'm doing it right now :dagger:

dpordomingo:
The first subtask is being addressed by https://github.com/src-d/gitbase-playground/issues/4 through https://github.com/src-d/gitbase-playground/pull/16
>_ FE: a `package.json` \w standard scripts that do `echo 1`_
so every other thigs seems to be completed.

PR #7, opened 2018-04-24
Backend skeleton

carlosms:
Part of #1.

The code is taken from the [code-annotation project](https://github.com/src-d/code-annotation).
It contains the bare minimum to start the server and serve `/version` and `/index.hml`.

The server can be started with:
```bash
$ GITBASEPG_ENV=dev go run cli/server/server.go 
INFO[2018-04-24 15:31:15] listening on 0.0.0.0:8080                    
```


bzz:
Nice! 

Shall we look into using https://github.com/src-d/go-log from the day 0 as well? There is no updated guideline, but it's a good chance to be an early adopters.

carlosms:
I think we follow all the rules in the conventions document.
Let's change to the new log lib in a separate PR.

Issue #2, opened 2018-04-13
Backend: http proxy to gitbase


bzz:
Implement 2 endpoints

1. Query: `POST /query`
```
req: { query:””, limit: N }
res: { meta:{..table headers..}, data:[row,...,], 
        status:{ elapsed, error/warning } 
}
```
No streaming yet. Include UAST -> JSON conversion.

2. Tables: `GET /tables`
```
res:  same
```

And two more endpoints as just a mocks

1.  UAST: `POST /parse`

   ```
	req: { code: “”, filter: “” }
	res: { uast: “”, status: {} }
   ```
    
2. Filter: `POST /filter`
```
  req: {} //need to include a binary PB
  res: { uast: “”, status {} }
```

PR #8, opened 2018-04-25
Query endpoint

carlosms:
Part of #2. Adds a new endpoint, `POST /query`. It forwards SQL queries to gitbase.

A few notes:

I've used the master branch as the constraint of `github.com/go-sql-driver/mysql` because the latest release, 1.3, does not support [ColumnType.DatabaseTypeName](https://golang.org/pkg/database/sql/#ColumnType.DatabaseTypeName). See https://github.com/go-sql-driver/mysql/pull/667

In the design we talked about adding extra information, such as the elapsed time, to the `meta` field. In this PR `meta` only contains the table headers. I think it's ok to add extra pieces of information later as we find the need for them.

~~I've left a couple of TODO comments with minor things, I think we can take care of that later on.~~ _-> Solved with the review suggestions_

.

The requests and results look like this:

```json
{
  "query": "SELECT name, hash, is_remote(name) AS is_remote FROM refs",
  "limit": 20
}
```

```json
{
    "status": 200,
    "data": [
        {
            "hash": "66fd81178abfa342f873df5ab66639cca43f5104",
            "is_remote": false,
            "name": "HEAD"
        },
        {
            "hash": "66fd81178abfa342f873df5ab66639cca43f5104",
            "is_remote": false,
            "name": "refs/heads/master"
        },
        {
            "hash": "66fd81178abfa342f873df5ab66639cca43f5104",
            "is_remote": true,
            "name": "refs/remotes/origin/master"
        }
    ],
    "meta": {
        "headers": [
            "name",
            "hash",
            "is_remote"
        ]
    }
}
```

A failure:
```json
{
	"query": "SELECT * FROM somewhere",
	"limit": 20
}
```

```json
{
    "status": 400,
    "errors": [
        {
            "status": 400,
            "title": "Error 1105: unknown error: table not found: somewhere"
        }
    ]
}
```

For a query with uast, the protobuf response is unmarshalled and the json is returned:
```json
{
	"query": "SELECT hash, content, uast(blobs.content, 'go') FROM blobs WHERE hash='fd30cea52792da5ece9156eea4022bdd87565633'",
	"limit": 20
}
```

```json
{
    "status": 200,
    "data": [
        {
            "content": "package server\n\nimport (\n\t\"net/http\"\n\n\t\"github.com/src-d/gitbase-playground/server/handler\"\n\n\t\"github.com/go-chi/chi\"\n\t\"github.com/go-chi/chi/middleware\"\n\t\"github.com/pressly/lg\"\n\t\"github.com/rs/cors\"\n\t\"github.com/sirupsen/logrus\"\n)\n\n// Router returns a Handler to serve the backend\nfunc Router(\n\tlogger *logrus.Logger,\n\tstatic *handler.Static,\n\tversion string,\n) http.Handler {\n\n\t// cors options\n\tcorsOptions := cors.Options{\n\t\tAllowedOrigins:   []string{\"*\"},\n\t\tAllowedMethods:   []string{\"GET\", \"POST\", \"PUT\", \"OPTIONS\"},\n\t\tAllowedHeaders:   []string{\"Location\", \"Authorization\", \"Content-Type\"},\n\t\tAllowCredentials: true,\n\t}\n\n\tr := chi.NewRouter()\n\n\tr.Use(middleware.Recoverer)\n\tr.Use(cors.New(corsOptions).Handler)\n\tr.Use(lg.RequestLogger(logger))\n\n\tr.Get(\"/version\", handler.APIHandlerFunc(handler.Version(version)))\n\n\tr.Get(\"/static/*\", static.ServeHTTP)\n\tr.Get(\"/*\", static.ServeHTTP)\n\n\treturn r\n}\n",
            "hash": "fd30cea52792da5ece9156eea4022bdd87565633",
            "uast(blobs.content, \"go\")": [
                {
                    "InternalType": "File",
                    "Properties": {
                        "Package": "1"
                    },
                    "Children": [
                        {
                            "InternalType": "Ident",
                            "Properties": {
                                "internalRole": "Name"
                            },
                            "Token": "server",
                            "StartPosition": {
                                "Offset": 9,
                                "Line": 1,
                                "Col": 10
                            },
                            "Roles": [
                                18,
                                1
                            ]
                        },
                        
                        [...]

                }
            ]
        }
    ],
    "meta": {
        "headers": [
            "hash",
            "content",
            "uast(blobs.content, \"go\")"
        ]
    }
}
```


server/handler/query.go
```diff
@@ -0,0 +1,88 @@
...
+	return columnValsPtr
+}
+
+// Query returns a function that forwards an SQL query to gitbase and returns
+// the rows as JSON
+func Query(db *sql.DB) RequestProcessFunc {
+	return func(r *http.Request) (*serializer.Response, error) {
+		var queryRequest queryRequest
+		body, err := ioutil.ReadAll(r.Body)
+		defer r.Body.Close()
```
smacker: https://golang.org/pkg/net/http/#Request
> The Server will close the request body. The ServeHTTP Handler does not need to.

server/handler/query.go
```diff
@@ -0,0 +1,88 @@
...
+			return nil, err
+		}
+
+		// TODO (carlosms) this only works if the query does not end in ;
+		// and does not have a limit. It will also fail for queries like
+		// DESCRIBE TABLE
+		query := fmt.Sprintf("%s LIMIT %d", queryRequest.Query, queryRequest.Limit)
+		rows, err := db.Query(query)
+		if err != nil {
+			return nil, serializer.NewHTTPError(http.StatusBadRequest, err.Error())
```
smacker: as TODO. Mysql returns `MySQLError` with 2 fields number&message. Would be nice to map them into our json.
https://github.com/go-sql-driver/mysql/blob/master/errors.go#L58

server/handler/query.go
```diff
@@ -0,0 +1,88 @@
...
+		columnValsPtr := genericVals(nColumns)
+
+		tableData := make([]map[string]string, 0)
+
+		for rows.Next() {
+			if err := rows.Scan(columnValsPtr...); err != nil {
+				return nil, err
+			}
+
+			colData := make(map[string]string)
```
smacker: `colData := make(map[string]string, nColumns)`

server/handler/query.go
```diff
@@ -0,0 +1,88 @@
...
+		query := fmt.Sprintf("%s LIMIT %d", queryRequest.Query, queryRequest.Limit)
+		rows, err := db.Query(query)
+		if err != nil {
+			return nil, serializer.NewHTTPError(http.StatusBadRequest, err.Error())
+		}
+		defer rows.Close()
+
+		columnNames, _ := rows.Columns()
+		nColumns := len(columnNames)
+		columnValsPtr := genericVals(nColumns)
```
smacker: how does it work for integers?

smacker: ah. I see you changed it in next commit.

server/handler/query.go
```diff
@@ -8,20 +8,34 @@ import (
...
-	columnValsPtr := make([]interface{}, nColumns)
-
-	for i := range columnVals {
-		columnValsPtr[i] = &columnVals[i]
+// genericVals returns a slice of interface{}, each one a pointer to the proper
+// type for each column
+func genericVals(colTypes []*sql.ColumnType) []interface{} {
+	columnValsPtr := make([]interface{}, len(colTypes))
+
+	for i, colType := range colTypes {
```
smacker: should we cover all types correctly? For example I don't see double or float.
https://github.com/go-sql-driver/mysql/blob/master/fields.go#L16

server/handler/query.go
```diff
@@ -0,0 +1,88 @@
...
+			err = json.Unmarshal(body, &queryRequest)
+		}
+
+		if err != nil {
+			return nil, err
+		}
+
+		// TODO (carlosms) this only works if the query does not end in ;
+		// and does not have a limit. It will also fail for queries like
+		// DESCRIBE TABLE
```
smacker: gitbase doesn't support anything besides SELECT, SHOW, DESCRIBE.
Add limit makes sense only to SELECT queries. Ending `;` also very easy to fix:
```
query = strings.TrimRight(strings.TrimSpace(query), ";")
```
not perfect but better than nothing

server/handler/query.go
```diff
@@ -0,0 +1,143 @@
...
+				case *sql.NullString:
+					sqlVal, _ := val.(*sql.NullString)
+					if sqlVal.Valid {
+						colData[columnNames[i]] = sqlVal.String
+					}
+				case *[]byte:
+					// TODO (carlosms) this may not be an array always
+					var protobufs [][]byte
+					if err := json.Unmarshal(*val.(*[]byte), &protobufs); err != nil {
+						return nil, err
```
dpordomingo: Should we return a 500 if one UAST cannot be read? I'd return the rows, and warning

server/handler/query.go
dpordomingo: +1

server/handler/query.go
```diff
@@ -52,28 +66,69 @@ func Query(db *sql.DB) RequestProcessFunc {
...
+					}
+
+					nodes := make([]*uast.Node, len(protobufs))
```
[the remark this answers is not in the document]
dpordomingo: +1

server/handler/query.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+
+			tableData = append(tableData, colData)
+		}
+
+		if err := rows.Err(); err != nil {
+			return nil, err
+		}
+
+		return serializer.NewQueryResponse(tableData, columnNames), nil
```
dpordomingo: won't be useful to send the type too? That way it won't be needed to check it in the frontend too.

server/handler/query.go
dpordomingo: +1

server/handler/query.go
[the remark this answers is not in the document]
dpordomingo: +1

server/handler/query.go
carlosms: Fixed, thanks! I actually took it from https://github.com/src-d/code-annotation/blob/master/server/handler/assignments.go#L82

server/handler/query.go
carlosms: Good idea, done in e6deab2

server/handler/query.go
carlosms: You are right, I just added the types I encountered them in gitbase replies, but it's better to make it complete since the beginning. Done in e6deab2

server/handler/query.go
```diff
@@ -52,28 +66,69 @@ func Query(db *sql.DB) RequestProcessFunc {
...
+					if sqlVal.Valid {
+						colData[columnNames[i]] = sqlVal.Int64
+					}
+				case *sql.NullString:
+					sqlVal, _ := val.(*sql.NullString)
+					if sqlVal.Valid {
+						colData[columnNames[i]] = sqlVal.String
+					}
+				case *[]byte:
+					// TODO (carlosms) this may not be an array always
```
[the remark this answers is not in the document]
carlosms: I was thinking that gitbase may at some point add something new that could return a JSON object. For instance, a new function 'first_uast_as_json()'

server/handler/query.go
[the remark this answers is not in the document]
carlosms: I've opened an issue so we don't forget about it when we start working on the integration with the uast frontend component:
https://github.com/src-d/gitbase-playground/issues/9

server/handler/query.go
carlosms: :+1: Implemented the todo with your suggestion in e6deab2

server/handler/query.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		if err != nil {
+			return nil, err
+		}
+
+		// TODO (carlosms) this only works if the query does not end in ;
+		// and does not have a limit. It will also fail for queries like
+		// DESCRIBE TABLE
+		query := fmt.Sprintf("%s LIMIT %d", queryRequest.Query, queryRequest.Limit)
```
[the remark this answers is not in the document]
carlosms: Because the functionality is so basic, for now I've moved it to a new function in e6deab2

server/handler/query.go
carlosms: I actually found a problem with this. gitbase can also return a JSON column with an array of strings, for example in [commits.parents](https://github.com/src-d/gitbase/blob/master/commits.go#L26)

With the changes in e6deab2 if the unmarshal fails the data will be returned as a string.

server/handler/query.go
carlosms: :+1: good suggestion. Done in e6deab2

server/handler/query.go
```diff
@@ -111,23 +113,21 @@ func Query(db *sql.DB) RequestProcessFunc {
...
+					// TODO (carlosms) this may not always be a JSON array, it could
+					// be a JSON object
+
+					// DatabaseTypeName JSON is used for arrays, but we don't know the
+					// type of each element. We try with uast node first and text later
+					nodes, err := unmarshallUAST(val)
+					if err == nil {
+						colData[columnNames[i]] = nodes
+					} else {
+						var strings []string
```
smacker: why strings? Maybe `[]interface{}` is better?

server/handler/query.go
```diff
@@ -138,6 +138,58 @@ func Query(db *sql.DB) RequestProcessFunc {
...
+
+	return nodes, nil
+}
+
+// addLimit adds LIMIT to the query, performing basic tests to skip it
+// for DESCRIBE TABLE, SHOW TABLES, and avoid '; limit'
+func addLimit(query string, limit int) string {
+	upper := strings.ToUpper(query)
+	if strings.Contains(upper, "DESCRIBE") || strings.Contains(upper, "SHOW") {
+		return query
```
smacker: please consider to rewrite it like this:

```go
func addLimit(query string, limit int) string {
  query = strings.TrimRight(strings.TrimSpace(query), ";")
  if strings.HasPrefix(strings.ToUpper(query), "SELECT") {
    return fmt.Sprintf("%s LIMIT %d", query, limit)
  }
  return query
```

it covers more cases and imo easier to read

server/handler/query.go
carlosms: I did it thinking in the commits.parents column, which is an array of strings. But it actually doesn't even have to be an array, using `interface{}` will work for any JSON array or object we receive. Changed in 28fc331.

server/handler/query.go
carlosms: Makes sense, changed in 28fc331

carlosms:
@bzz, a comment for your consideration when adding your review.
We discussed via chat and we agree that since we are still at this very early dev stage we can do tests in a future PR.

bzz:
@carlosms as it seems to be almost ready, do you think at this point it would be reasonable to ask to squash commits before a final pass? `+37,039 −11` in 126 files looks a bit harsh and all previous feedback seems to be already 👍

cmd/server/main.go
```diff
@@ -1,25 +1,33 @@
...
 var version = "dev"
 
+// Note: maxAllowedPacket must be explicitly set for go-sql-driver/mysql v1.3.
+// Otherwise gitbase will be asked for the max_allowed_packet column and the
+// query will fail.
+// The next release should make this parameter optional for us:
+// https://github.com/go-sql-driver/mysql/pull/680
 type appConfig struct {
 	Env       string `envconfig:"ENV" default:"production"`
 	Host      string `envconfig:"HOST" default:"0.0.0.0"`
```
bzz: Sorry if I missed that somehow, not very related to the change, but while we are here - did not we have issues with this before, being not a valid IP on some OSes and decided that default should be localhost or 127.0.0.1 instead?

Remember discussing it with @smacker

server/handler/query.go
```diff
@@ -0,0 +1,192 @@
...
+func Query(db *sql.DB) RequestProcessFunc {
+	return func(r *http.Request) (*serializer.Response, error) {
+		var queryRequest queryRequest
+		body, err := ioutil.ReadAll(r.Body)
+		if err == nil {
+			err = json.Unmarshal(body, &queryRequest)
+		}
+
+		if err != nil {
+			return nil, err
```
bzz: Question on user input validation on API level - what is the expected behaviour of playground on `/query` endpoint, in case `"limit":"20"` is specified, in payload, instead of `"limit":20` ?

As an API user, I would expect it to result in
```json
{
    "status": 4XX,
    "errors": [
        {
            "status": 4XX,
            "title": "Error description"
        }
    ]
}
```

but I belive current implementation would result in 500 without any payload in response.

carlosms: Improved in 99af6d0. Now limit is optional, and if the json unmarshall fails the reply will contain:
```json
{
    "status": 400,
    "errors": [
        {
            "status": 400,
            "title": "Bad Request. Expected body: { \"query\": \"SQL statement\", \"limit\": 1234 }"
        }
    ]
}
```

carlosms:
@bzz:
> As it seems to be almost ready, do you think at this point it would be reasonable to ask to squash commits before a final pass?

I was planning to squash before the merge, but left the small commits so you could follow the PR discussions easier.

I've squashed the previous commits and added 2 small new ones. Please take another look and I'll squash again before merging.

> create a /docs/rest-api.md with a single section on Query, and a paragraph that includes those few CURL commands.

Added in 7dc93b3.

cmd/server/main.go
dpordomingo: is not `0.0.0.0` the right one?
That's how it was configured in CAT, because that way the container is listening "everywhere"
https://github.com/src-d/code-annotation/pull/153/files

smacker: `0.0.0.0` is a correct one. We use this variable for the server. And server should listen on all IPs.

smacker:
@bzz sorry, I don't remember such requirement in the design document. I also don't understand how it would work. Should it open 2 tabs? We didn't discuss it at all.

carlosms:
I also don't remember any mention to multiple queries :thinking:

bzz:
@carlosms @dpordomingo not sure, what do you guys suggest on how to proceed here, but I'm totally open for discussing/updating design doc next week on "multiple query user input" feature, potentially handing it in subsequent PRs and merging this "as is" for now. 

Please let me know!

dpordomingo:
Is gitbase prepared to accept two queries, or does it return an error?
From @carlosms message it seems that it returns a syntax error, so if playground should not parse nor pre validate the user input, three current behavior seems to follow the agreed expectations, doesn't it?

bzz:
> Is gitbase prepared to accept two queries, or does it return an error?

sorry, was not able to test before https://github.com/src-d/gitbase/issues/255 but may be @carlosms could clarify?

> .. behavior seems to follow the agreed expectations, doesn't it?

I think it does that way

carlosms:
> > Is gitbase prepared to accept two queries, or does it return an error?
> 
> sorry, was not able to test before src-d/gitbase#255 but may be @carlosms could clarify?

It will result in error. If we need to support multiple queries separated by `;` we would need to separate the strings either in the backend or the frontend.

And I still think this would not be the expected behaviour for most people. One query -> one tab; one request -> one response; anything else would be confusing.

bzz:
> > > Is gitbase prepared to accept two queries, or does it return an error?
> > 
> > sorry, was not able to test before src-d/gitbase#255 but may be @carlosms could clarify?
> It will result in error

:+1: this is a strong argument for merging this work asap and iterate lated on, based on product feedback (as noted by @dpordomingo it follows the conventions we've agreed up on).

PR #10, opened 2018-04-27
Tests for query handler

carlosms:
Part of #2.
Depends on #8, the only new commit is c58000f.

The tests require a working gitbase server and bblfshd server. gitbase should be serving at least this repo (https://github.com/src-d/gitbase-playground).

This will be added to the travis conf in the future, but meanwhile it can be done manually:

```bash
docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
docker exec -it bblfshd bblfshctl driver install --all

go run cli/gitbase/main.go server --git=repos -v
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,49 @@
+## Development
+
+### Dependencies
+
+Launch [bblfshd](https://github.com/bblfsh/bblfshd) and install the drivers. More info in the [bblfshd documentation](https://doc.bblf.sh/user/getting-started.html).
+
+```bash
+docker run -d --name bblfshd --privileged -p 9432:9432 -v /var/lib/bblfshd:/var/lib/bblfshd bblfsh/bblfshd
+docker exec -it bblfshd bblfshctl driver install --all
+```
+
+Install [gitbase](https://github.com/src-d/gitbase), populate a repository directory, and start it.
+
+```bash
+go get github.com/src-d/gitbase/...
+cd $GOPATH/src/github.com/src-d/gitbase
+make dependencies
+mkdir repos
+git clone https://github.com/src-d/gitbase-playground.git repos/gitbase-playground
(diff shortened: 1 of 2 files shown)


carlosms:
@bzz 

> @carlosms could you help me understand, is this a set integration tests that would make `go test` fail if no bblfsh/gitbase is running and functionally equivalent to some series of CURL commands?

Yes, exactly.

> Would it be possible to separate tests, that depend on external infrastructure (even local) and have them behind a flag, off by default? In my experience, from developer productivity standpoint, it's usually is very nice when the default way of running tests "just runs" quickly.

Honestly I wouldn't go into this now, we don't even have a makefile. And so far these integration tests are the only ones we have. When we have more tests I think we could have two separate `make` commands for local and integration tests.

Let me know if you have something else in mind that can be done now.

> Another thing is - as development flow becomes more complex and to make the PR more self-contained, could you please also include a `REAMDE.md` file with a single section `Development` like
> 
> ```
> ## Development
> ### Build
> <commands how to build>
> ### Run
> <commands how to run, assuming it's built>
> ### Run tests
> <dependencies> (with links i.e [run Bblfsh](https://doc.bblf.sh/user/getting-started.html#running-with-docker-recommended))
> <commands how run tests>
> ```
> 
> No need to wary about doc completeness, applying guidelines and explaining anything in english - at this point just a series of CLI commands would be more then enough and later this minimal work will refactored and find it's appropriate place in the overall project's documentation.

Done in f07c127.

carlosms:
> although for me, having all tests that require external dependency behind the flag is the expected default behaviour, up to @carlosms - if that is something you want to add later, sounds good.

Merging. Let's keep that in mind when we work on CI.

PR #13, opened 2018-05-07
Project structure

dpordomingo:
required by https://github.com/src-d/gitbase-playground/issues/1
blocks https://github.com/src-d/gitbase-playground/pull/14

- Add MAINTAINERS, .gitignore and ,editorconfig files 2d8654c 4e500f0 301e291
- Put frontend code under 'public' directory 491fdd5
- Rename playground executable 51e132a
- Enhance Gopkg.toml, update missing dependencies and prune unnecessary stuff  d2b755b 5ef69d8 8ad6c63 b96ded1


Gopkg.lock
```diff
@@ -50,6 +56,12 @@
   revision = "645ef00459ed84a119197bfb8d8205042c6df63d"
   version = "v0.8.0"
 
+[[projects]]
+  name = "github.com/pmezard/go-difflib"
```
bzz: Are we going to use diffs in gitbase playground?

bzz: Wow!.. thank you for research and a followup

bzz:
\cc @carlosms for a quick glance

cmd/gitbase-playground/main.go
```diff
@@ -48,7 +48,7 @@ func main() {
 	}
 	defer db.Close()
 
-	static := handler.NewStatic("build", conf.ServerURL)
+	static := handler.NewStatic("build/public", conf.ServerURL)
```
carlosms: I doesn't matter that much because this will change when the frontend stuff is added. But shouldn't this be pointing to `public` now that you moved `index.html` there?

dpordomingo: nope; as I understood, `public` will contain frontend sources; these front sources will be compiled by `yarn` (next iteration) and the generated code will be stored in `build/public` (to be read by `bindata`)
The logic in behind is to keep the process done in `code-annotation`, but with one extra level depth

PR #14, opened 2018-05-07
Add some tests to be used by the CI

dpordomingo:
required by #1

Added tests:
- 345e890 It was added some integration tests to ensure that `/version` endpoint works,
- fe788ec It was added empty tests to get a proper coverage report,

Others:
- 86e1d6b To make the app easily testable it was used an interface of `*sql.DB`

### DIFF
--- a/server/common_test.go
+++ b/server/common_test.go
@@ -1,0 +1,59 @@
+package server_test
+
+import (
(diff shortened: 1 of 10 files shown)


bzz:
\cc @carlosms for a quick look

PR #15, opened 2018-05-08
Add CI, docs and fix integration tests

dpordomingo:
required by #1 
address #5

I ran it in my own Travis, in a private PR https://github.com/dpordomingo/gitbase-playground/pull/8 and here is the result:
- on `pull_request` &rarr; PR review [stage 1](https://travis-ci.org/dpordomingo/gitbase-playground/builds/376331133)
- on merge agains `master` &rarr; integration review [stage 1](https://travis-ci.org/dpordomingo/gitbase-playground/builds/376332730)
- on `tag` &rarr; release docker images and GH assets [stage 1+2+3](https://travis-ci.org/dpordomingo/gitbase-playground/builds/376334429)

Tag with assets to validate (darwin and linux) &rarr; [dpordomingo/gitbase-playground#v0.0.1](https://github.com/dpordomingo/gitbase-playground/releases/tag/v0.0.1)

Docker image to validate &rarr; [hub.docker.com/dpordomingo/gitbase-playground](https://hub.docker.com/r/dpordomingo/gitbase-playground/tags/)

### Assumptions for the first iteration

- it's used `dpordomingo` accounts for DockerHub and Travis till there are company ones,
- it's used `dpordomingo/ci` because there are PR in `src-d/ci` that has not been merged yet,<br /> (https://github.com/src-d/ci/pull/67 and https://github.com/src-d/ci/pull/62)
- there is no frontend yet, but the `make` rules are copied from `code-annotation` (it might change in the next iteration),

# Documentation

It was added:
- `README.md`
- `docs/`
  - `CONTRIBUTING.md`
  - `quickstart.md` guideline to run the playground with its dependencies
- `CODE_OF_CONDUCT.md` and `DCO`

# CI in Travis:

- There are three stages (each one of them runs if the prev one succeeded):
  - `validate and build` to ensure the QA, that runs on `pull_request`, `tag`, `master` and `staging`,
  - `release to github` to publish build assets (linux/darwin) to GH release on every `tag`,
  - `push to dockerhub` to publish the docker image (+ `:latest`) to DockerHub on every `tag`,
- Travis conf should use as much `src-d/ci::v1` api as possible,
- `make validate-commit` checks that:
  - there is no fixable/lintable code,
  - no vendors were missing, and 
  - `./server/assets/asset.go` proxy was not ruined, 

# Integration tests

[...]

Issue #4, opened 2018-04-13
Frontend: implement simple UI


bzz:
Implement main functional components:
 - query
 - run button
 - result table (no any pagination is ok at this stage)

take care of passing env var to fronted. 
At this point, query should always add LIMIT N with N=100 defined though env var.

carlosms:
Ready in master :tada:

PR #16, opened 2018-05-08
Simple UI

carlosms:
Part of #4.

localhost_3000_

This is a simple first version for the UI. A few notes:

- To keep this PR smaller and focused on the frontend, the following part of the issue description will be done in a different PR:
> take care of passing env var to fronted.
> At this point, query should always add LIMIT N with N=100 defined though env var.

- I didn't include instructions in the readme because it will change a lot once #15 is merged. To test this branch start the go server and then run 
```
cd frontend
yarn start
```

- In @dpordomingo's code he is assuming the new frontend code would live in `/public`. But I find it a bit strange, because the fronted actually has a sub-dir named `public` for some files. I this PR I propose `/frontend`, but this can be easily changed.
- We discussed about adding redux to prepare for future increasing complexity. The current code is so simple that I preferred to go without it for now. I would suggest to add it once we add several result tabs. But if you think it should be included now in this PR or a new one, let's discuss it.
- The styling is so minimalistic that there is none :D
- I chose [react-table](https://github.com/react-tools/react-table) for the results table. As you can see the code is quite simple and it can easily changed to any other option should we need to. Let me know if you are aware of any other better lib.
- I didn't forget about the debug text area at the bottom. I left it because it can be quite useful at this stage of the project, and the code is very localized and easy to remove later.


frontend/src/App.js
```diff
@@ -44,6 +44,13 @@ GROUP BY committer_email, month, repo_id`,
   }
 
   render() {
+    let ok;
```
smacker: wow. why not
```js
const ok = this.state.response && this.state.response.status === 200
```
?

try-catch isn't very javascript way to do such things.

or even better:
```js
const { response } = this.state;
const ok = response && response.status === 200;
```

smacker:
btw do I understand right that you are going to implement proper layout, loading and error state in separate PRs? According to the issue, I would expect it.

carlosms:
@smacker:
> btw do I understand right that you are going to implement proper layout, loading and error state in separate PRs? According to the issue, I would expect it.

Of course that's needed, but my understanding from the design doc is that issue #4 (alpha 1 milestone) would only include the most basic UI (as is in this PR), and those things you list would be added for alpha 2, new issues.

ping @bzz for clarification.

bzz:
For me, it would be totally fine to expect proper error placement and further enhancement of current components as part of the alpha2 as Carlos suggests, for the reason of getting a really minimal alpha1 faster to the staging env.

bzz:
GUI looks great.

BTW if we already have table, may be we do not need to print JSON

frontend/.eslintrc.js
```diff
@@ -0,0 +1,25 @@
...
+  ],
+  env: {
+    browser: true,
+    es6: true,
+    node: true,
+    'jest/globals': true,
+  },
+  plugins: ['import', 'react', 'jest', 'prettier'],
+  rules: {
+    'prettier/prettier': 'error',
```
dpordomingo: comparing with `cat`, why not adding?
```yml
'prettier/prettier': ['error', { singleQuote: true, trailingComma: 'es5' }],
```

frontend/.eslintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    'import/no-extraneous-dependencies': ['error', { 'devDependencies': ['**/*.test.js', 'src/setupTests.js'] }],
+    'import/no-unresolved': 0,
+    'import/extensions': 0,
+    'func-names': 0,
+    'no-plusplus': ['error', { allowForLoopAfterthoughts: true }],
```
dpordomingo: I miss other rules from `cat`
```yml
'no-underscore-dangle': 0, // because of _super
'react/prop-types': 0,
'no-console': ['error', { allow: ['info', 'warn', 'error'] }], // later we will might need wrapper
```
why not adding them?

frontend/.gitignore
```diff
@@ -0,0 +1,21 @@
+# See https://help.github.com/ignore-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
+
+# testing
+/coverage
```
dpordomingo: really needed for the front? it's already in the main `.gitignore`

frontend/.gitignore
```diff
@@ -0,0 +1,21 @@
...
+
+# production
+/build
```
dpordomingo: I think it's not needed by the front... it's already in main `.gitignore`

frontend/.gitignore
```diff
@@ -0,0 +1,21 @@
...
+
+# misc
+.DS_Store
```
dpordomingo: all this `misc` ones, with `.DS_Store` and `.env.*` are already in the main `.gitignore`

frontend/.gitignore
```diff
@@ -0,0 +1,21 @@
...
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
```
dpordomingo: in some cases I needed to add `.eslintcache` :D

frontend/.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
+  "singleQuote": true
```
dpordomingo: I miss from `cat`
```yaml
"trailingComma": "es5"
```

frontend/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "react-helmet": "^5.2.0",
+    "react-scripts": "1.1.4",
+    "react-table": "^6.8.2"
+  },
+  "scripts": {
+    "start": "react-app-rewired start",
+    "build": "react-app-rewired build",
+    "test": "react-app-rewired test --env=jsdom",
+    "eject": "react-scripts eject",
+    "lint": "eslint 'src/**/*.js'",
```
dpordomingo: in `cat` we also run `prettier` when linting
```yml
    "lint": "eslint 'src/**/*.js' && prettier -l 'src/**/*.js'",
```

frontend/public/index.html
```diff
@@ -0,0 +1,40 @@
...
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <meta name="theme-color" content="#000000">
+    <!--
+      manifest.json provides metadata used when your web app is added to the
+      homescreen on Android. See https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/
+    -->
+    <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
+    <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
```
dpordomingo: what if we use the `cat` favicon instead of the react one?

frontend/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "React App",
```
dpordomingo: we should use the gitbase-playground values :dancer:

frontend/src/index.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import 'bootstrap/dist/css/bootstrap.css';
+import 'bootstrap/dist/css/bootstrap-theme.css';
+import './override.less';
+import './index.css';
```
dpordomingo: could we write only `.less` code, as in `cat`?

frontend/src/api.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  return fetch(url, fetchOptions)
+    .then(checkStatus)
+    .then(resp => resp.json())
+    .then(json => {
+      if (json.errors) {
+        throw json.errors;
+      }
+      return json;
```
dpordomingo: the data will be always under a `data` field; why not returning it as done in `cat`?

frontend/src/api.js
```diff
@@ -0,0 +1,82 @@
+function checkStatus(resp) {
```
dpordomingo: why not moving this service into its own place?
In `cat` we used `/api` package, for example.

frontend/.gitignore
carlosms: The gitignore is the create-react-app default. If we see this dir as a kind of sub-project I think it makes sense to leave this `.gitignore` here for frontend-only files.

Plust the rules in the `.gitignore` in master do not apply here, instead of `/node_modules` it should be `frontend/node_modules`, right?

frontend/.gitignore
carlosms: Let's please keep the conversation in one thread, above.

frontend/.prettierrc
carlosms: Please see the conversation in https://github.com/src-d/company/pull/127

frontend/package.json
carlosms: If I'm not mistaken this is not needed when we use `eslint-plugin-prettier`.

frontend/src/index.js
carlosms: I would skip this and merge as it is. All the machinery to write `less` is in place, this file can be changed at any moment when we start to write styles.

frontend/src/api.js
carlosms: Not really, for this project we have `data` and `meta`.

frontend/src/api.js
carlosms: What would be the benefit of moving `api.js` to `/api/index.js`? I don't have a strong opinion on this, let me know why and we can move it.

frontend/public/index.html
carlosms: Done in 3d57d81.

frontend/public/manifest.json
carlosms: Done in 3d57d81.

frontend/src/index.js
carlosms: Done in 3d57d81.

carlosms:
@dpordomingo please take another look when you have time.

frontend/.eslintrc.js
dpordomingo: I didn't found agreement about 'no-console' conf; did you?
Are we accepting `console` things?
Why not using a [wrapper as in  `CAT`](https://github.com/src-d/code-annotation/blob/master/src/services/log.js)?

frontend/.gitignore
dpordomingo: I'm not sure if the folder is like a "subproject"...

I use to like small `.gitignore` files... with as fewer rules as possible, so I like the idea of avoiding duplicities; anyhow this is more like an opinion, so if you do not share it, feel free to disregard it :dagger: 

About `node_modules` rule:
if it should be in the main `.gitignore` it should be `frontend/node_modules` but since it makes sense to be in the `frontend/.gitignore` it needs to relative to the `frontend` dir.

frontend/src/api.js
dpordomingo: I see two reasons;

first one is kindof opinionated: imo the root dir should be populated with as fewer things as possible because it helps to understand better the structure of the code.

second one is generic: now we have in this file the logic (copy-pasted from `cat`) to connect with the API (business agnostic), and also the "repository-like" code (business related). Splitting both parts properly, and moving business agnostic to a service, and the business logic to a repository would help us to reuse and test better.

carlosms:
Thank you for your comments. I agree, those details can be discussed in new issues and changed later on without much difficulty. Please fell free to open any new issues.

Let's merge so we can move forward with the building and release tasks.

dpordomingo:
thanks!
issues created https://github.com/src-d/gitbase-playground/issues/28 and https://github.com/src-d/gitbase-playground/issues/29

PR #17, opened 2018-05-08
 Pass env vars from backend to frontend

carlosms:
Part of #4, depends on #16.
The only new commit is cec12d4.

It uses the same mechanism as code-annotation.

### DIFF
--- a/cmd/gitbase-playground/main.go
+++ b/cmd/gitbase-playground/main.go
@@ -23,20 +23,18 @@ var version = "dev"
 // The next release should make this parameter optional for us:
 // https://github.com/go-sql-driver/mysql/pull/680
 type appConfig struct {
-	Env       string `envconfig:"ENV" default:"production"`
-	Host      string `envconfig:"HOST" default:"0.0.0.0"`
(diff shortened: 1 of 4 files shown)


carlosms:
@smacker:
> how does it work with yarn start? Frontend & backend are on different ports, aren't they? Or do I miss something?

Sorry, I should have explained this part in the issue description, it works a bit different to code-annotation.

Following the [create-react-app docs](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#proxying-api-requests-in-development) recommendation, [`package.json`](https://github.com/src-d/gitbase-playground/pull/17/files#diff-0a6694dd0d6b93a2aa14f32e3d873b4cR34) contains `"proxy": "http://localhost:8080"`. 

By default `SERVER_URL` is empty, so for development the api call to `/query` is sent to webpack (localhost:3000) and proxied to the go backend (localhost:8080). 

For production, with the default empty `SERVER_URL` the api calls are still relative to the URL used to access the backend, so it will still work with no problem. Optionally, the env var can be set to force all api calls to a specific URL.

smacker:
@carlosms great!

PR #26, opened 2018-05-09
Basic UI improvements

carlosms:
Part of #4, depends on #16 & #17.
The only new commit is f0c9249.

Adds basic layout and style improvements:
- Remove debug box
- Show error messages
- Show loading animation and disable RUN button while loading.

a

-------------

b
-------------
c
-------------
d

Issue #18, opened 2018-05-09
Add tabs


carlosms:
Each query request opens a new tab containing:
- the SQL query string
- a link or button "edit query" that fills the main query box with this tab's query
- the results table

Ideally the top and bottom areas (query and tabs) can be resized with a drag-able separator.

PR #27, opened 2018-05-10
Frontend: Add tabs

### LINKED ISSUE #21: Frontend: display errors in the results section
Display any error messages where the results table would appear.

This may change to follow the wireframes later on.

carlosms:
Part of #18, also fixes #21.
The only thing missing from #18 is the split panels resize.

Depends on #26, the only new commit is 16db3da.

peek 2018-05-10 20-06


smacker:
I see a flaw in the implementation. It based on the fact that only one query would be run at one moment of time. But I definitely remember we discussed that playground should support multiple queries.

smacker:
the follow up to the offline discussion about setState. It's better to do like this when you need something like global counter:
```js
let globalCounter = 0;
class MyComponent extends Components {
  handler() {
    const c = ++globalCounter;
    // no race condition anymore
  }
}
```

carlosms:
Yes, I saw that potential problem with the async setState, but decided to improve it when we handle the inclusion of redux.

As you suggest, the fix is quite small so I pushed a new commit with your code.

frontend/src/App.js
```diff
@@ -12,66 +11,80 @@ class App extends Component {
...
   }
 
   handleSubmit() {
-    this.setState({
-      response: undefined,
-      loading: true,
-      error: undefined
-    });
+    const { sql } = this.state;
+    const key = ++this.uniqueKey;  // eslint-disable-line
```
dpordomingo: I'd avoid exceptions and allow it in `.eslintrc.js`

frontend/src/App.js
```diff
@@ -12,66 +11,80 @@ class App extends Component {
...
       })
       .catch(msgArr => {
-        this.setState({ loading: false, error: msgArr.join('; ') });
+        if (!this.state.results.has(key)) {
+          // Tab was removed, ignore results
+          return;
+        }
+
+        const newResults = new Map(this.state.results);
+        newResults.set(key, { sql, errorMsg: msgArr.join('; ') });
```
dpordomingo: `then` and `catch` is almost the same but for one part of this line... I'd try to refactor...

frontend/src/components/TabbedResults.less
```diff
@@ -0,0 +1,40 @@
+.close {
+  width: auto;
```
dpordomingo: tab indent should be with [4 spaces](https://github.com/carlosms/gitbase-playground/blob/b55a986dd66ad8e642f5ccd90a21d277b4c7386f/.editorconfig#L13)...
We should open an issue to address this inconsistency...

carlosms: I created a new issue: https://github.com/src-d/gitbase-playground/issues/37

frontend/src/App.js
carlosms: done in 71c1404

frontend/src/App.js
carlosms: done in 71c1404

PR #31, opened 2018-05-16
Fix static 404

dpordomingo:
Nonexistent files under `/static` folder will be served as `404` instead of serving `/index.html`

### DIFF
--- a/server/handler/static_files.go
+++ b/server/handler/static_files.go
@@ -4,8 +4,17 @@ import (
 	"bytes"
 	"encoding/json"
 	"net/http"
+	"path"
+	"strings"
 
 	"github.com/src-d/gitbase-playground/server/assets"
+)
+
+const (
+	staticDirName = "static"
+	indexFileName = "index.html"
+
+	serverValuesPlaceholder = "window.REPLACE_BY_SERVER"
 )
 
 // Static contains handlers to serve static using go-bindata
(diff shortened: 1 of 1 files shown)


server/handler/static_files.go
```diff
@@ -33,19 +35,25 @@ type options struct {
 
 // ServeHTTP serves any static file from static directory or fallbacks on index.hml
 func (s *Static) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	filepath := s.dir + r.URL.Path
+	filepath := path.Join(s.dir, r.URL.Path)
 	b, err := assets.Asset(filepath)
 	if err != nil {
+		if strings.HasPrefix(filepath, path.Join(s.dir, "static")) {
```
smacker: maybe it's better to avoid hard-coded directory name here and put in constant?

smacker:
👍 for the change. But code coverage is down not sure if it's a problem.

server/handler/static_files.go
dpordomingo: This file contains other two strings that should be extracted to constants:
- `window.REPLACE_BY_SERVER`
- `index.html`
For consistency, I'll define all the three constants if you agree @smacker

dpordomingo:
About codecov I created a separated issue for that https://github.com/src-d/gitbase-playground/issues/32
imo, since we've not proper tests here, we should not focus in that in this PR

server/handler/static_files.go
smacker: that would be great, thanks!

dpordomingo: done in the rebased/amended commit, thanks

dpordomingo:
@carlosms , as described in the PR desc, nonexistent files under `/static` are being served as `index.html` what is wrong; they should be served as `404` instead

carlosms:
Sounds good. I was just wondering if our code was requesting non-existing files.

PR #35, opened 2018-05-17
Add CodeMirror editor with sql syntax and autocomplete

### LINKED ISSUE #20: Frontend: SQL syntax highlight
Add SQL syntax highlight using the same mechanism as [babelfish dashboard](https://github.com/bblfsh/dashboard).

carlosms:
Fixes #20.

Besides syntax highlighting this PR also adds autocomplete. To do that, the schema is loaded when the application starts. Also the query can be executed with ctrl+enter.

peek 2018-05-17 19-01

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -87,6 +87,42 @@ function query(sql) {
   });
 }
 
+function tables() {
+  return apiCall(`/tables`);
+}
+
+/* Returns an array in the form:
+[
+  {
+    "table": "refs",
+    "columns": [
+      {
+        "name": "repository_id",
+        "type": "TEXT"
+      },
+      ...
+    ]
+  },
+  ...
+]
+*/
+function schema() {
+  return tables()
+    .then(res =>
+      Promise.all(
+        res.data.map(e =>
+          query(`DESCRIBE TABLE ${e.table}`).then(tableRes => ({
+            table: e.table,
+            columns: tableRes.data
+          }))
+        )
+      )
+    )
+    .catch(err => Promise.reject(normalizeErrors(err)));
+}
+
 export default {
-  query
+  query,
+  tables,
+  schema
 };
--- a/frontend/src/components/QueryBox.js
+++ b/frontend/src/components/QueryBox.js
@@ -1,20 +1,82 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { Row, Col, Button } from 'react-bootstrap';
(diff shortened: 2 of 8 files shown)


frontend/src/components/QueryBox.js
```diff
@@ -1,20 +1,60 @@
...
+      {}
+    );
+  }
+
   render() {
+    const tables = this.props.schema
+      ? this.schemaToCodeMirror(this.props.schema)
+      : {};
+
+    const options = {
```
smacker: Isn't really important, but I wouldn't convert schema and create options object on every render (means every key press). Consider to create the codemirror configuration in constructor and update it on componentWillReceiveProps when schema or `handleSubmit` are changed.

frontend/src/components/QueryBox.less
```diff
@@ -2,11 +2,14 @@
...
 
-button {
-  width: 100%;
+.CodeMirror {
+  height: 150px;
 }
+
+.CodeMirror-empty {
+  color: dimgrey;
+}
```
smacker: new line

frontend/src/components/QueryBox.js
carlosms: You are right, fixed in d2fb518

carlosms: rebased, now it's fixed in https://github.com/src-d/gitbase-playground/pull/35/commits/ce8f1dd8d7649dd3a4b55e822dfc0f7453ac6506

frontend/src/api.js
```diff
@@ -87,6 +87,42 @@ function query(sql) {
...
+  },
+  ...
+]
+*/
+function schema() {
+  return tables()
+    .then(res =>
+      Promise.all(
+        res.data.map(e =>
+          query(`DESCRIBE TABLE ${e.table}`).then(tableRes => ({
```
bzz: @carlosms sorry for digging this up, but I think we need to change this logic eventually.

Instead for 
 - API call /tables
 - then, for each table call DESCRIBE TABLE from the frontend

We should have
 - API call to /tables, so that the backend returns all this information

as that was one of the purposes of having separate API endpoint on the backend: to abstract the details of how meta-information (schema) is collected.

carlosms: I agree with you comment. I actually though of doing it that way in this PR, but didn't want to get too ahead of myself and went for the quicker approach of doing it in the frontend.

I've created a new issue with your request: https://github.com/src-d/gitbase-playground/issues/40

PR #38, opened 2018-05-21
Split panels vertically with resize

### LINKED ISSUE #18: Add tabs

carlosms:
Fixes #18.

Splits panels vertically with resize.

image


frontend/src/App.less
```diff
@@ -1,12 +1,50 @@
+@import '../node_modules/bootstrap/less/variables.less';
+
 .app {
-  margin-top: 3em;
-  margin-bottom: 3em;
+    padding-top: 3em;
+    padding-bottom: 1em;
+    height: 100vh;
```
smacker: maybe I'm too old school, but I would prefer

```css
html, body {
  margin: 0;
  padding: 0;
  height: 100%;
}

.app {
  height: 100%;
}
```

it's just easier to think about. Plus `vh` isn't accepted standart yet (even well supported across all browsers)

carlosms:
@smacker thank you for your comments. I tried to solve the issues you mention, please give it another look.

smacker:
> The pager must be always visible and the table should have scroll instead.

Fixed in Chrome, but doesn't work in Firefox.

carlosms:
> Fixed in Chrome, but doesn't work in Firefox.
Should be fixed now. I hope.

bzz:
Nice work! Was a bit confused with the description `Splits panels vertically ...` at first.

If I'm not mistaken, it looks like a horizontal split on the screenshot.

carlosms:
> it looks like a horizontal split on the screenshot

I never said I was good at this design stuff, did I? :smile:  Did not realize the mistake, don't worry the code is splitting the right way.

> do you think it is worth to update a screenshot link in README

Yes, we should keep that screenshot updated. I was thinking of doing it for each release.

frontend/src/App.less
```diff
@@ -1,12 +1,62 @@
...
+    margin-top: 3em;
+}
+
+// react-split-pane
+.SplitPane .Pane {
+    overflow: auto;
+}
+
+.SplitPane {
+    position: relative !important;
```
dpordomingo: why not nesting here `.Pane` and `.Resizer`?

frontend/src/components/TabbedResults.less
```diff
@@ -1,36 +1,60 @@
...
-.tab-content > .tab-pane {
-  margin-top: 2em;
+#tabbed-results {
+    display: flex;
+    flex-direction: column;
+}
+
+.tab-content {
+    flex: auto 1;
+
```
dpordomingo: I'd group `flex` rules and remove the extra empty line.

carlosms: This is both a flex item (`flex`) and a flex container (`display` & `flex-direction`), I think it makes sense to have them separated.

Issue #32, opened 2018-05-16
Test handlers


dpordomingo:
Code coverage is really bad...

carlosms:
Closing because we have the infrastructure for unit tests now. Although we should definitely continue creating more tests.

PR #41, opened 2018-05-22
Add integration tests to Travis CI

carlosms:
Part of #32.

This PRs adds the already implemented integration tests to Travis CI.

The docker image `carlosms/bblfsh-ci-test:v0.0.1` contains the latest bblfshd image plus the recommended drivers (created with the [instructions here](https://github.com/bzz/ml-on-code/blob/master/images/README.md#create-bblfshd--drivers-image)). Before merging it would be better to host this image in the srcd organization.

It looks like the default gitbase port 3306 is in use in the travis image, that's why I'm using a different one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,16 @@ jobs:
         - make coverage
         - make validate-commit
     - script:
+      - docker run -d --name bblfshd --privileged --volume $HOME/bblfsh-drivers:/var/lib/bblfshd -p "9432:9432" bblfsh/bblfshd
+      - sleep 5
+      - docker exec -it bblfshd bblfshctl driver install go docker://bblfsh/go-driver:latest
+      - docker exec -it bblfshd bblfshctl driver install javascript docker://bblfsh/javascript-driver:latest
+      - mkdir $HOME/repos
+      - git clone https://github.com/src-d/gitbase-playground.git $HOME/repos/gitbase-playground
+      - docker run -d --name gitbase -p "3367:3306" -e "BBLFSH_ENDPOINT=bblfshd:9432" --volume $HOME/repos:/opt/repos --link bblfshd srcd/gitbase
+      - sleep 15
+      - GITBASEPG_DB_CONNECTION='gitbase@tcp(localhost:3367)/none' make back-test-integration
+    - script:
         - make packages
     # release to github
     - stage: release to github
@@ -60,3 +70,11 @@ jobs:
       script:
         - PKG_OS=linux make build
         - DOCKER_PUSH_LATEST=true make docker-push
+
+cache:
+  directories:
+    - $HOME/bblfsh-drivers/images
+
+before_cache:
+  # make bblfsh images readable
+  - sudo chmod -R 777 $HOME/bblfsh-drivers/images
--- a/server/handler/common_test.go
+++ b/server/handler/common_test.go
@@ -51,7 +51,7 @@ func (suite *HandlerSuite) TearDownSuite() {
 }
 
 type appConfig struct {
-	DBConn string `envconfig:"DB_CONNECTION" default:"root@tcp(localhost:3306)/none?maxAllowedPacket=4194304"`
+	DBConn string `envconfig:"DB_CONNECTION" default:"gitbase@tcp(localhost:3306)/none?maxAllowedPacket=4194304"`
 }
 
 func getDB(isIntegration bool) (service.SQLDB, error) {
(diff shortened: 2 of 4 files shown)


carlosms:
I've pushed a new commit with a dockerfile to create this modified bblfsh image with the drivers installed.

@bzz how should we proceed with this new docker image? Where should we host it? Do we need to involve the Infrastructure team?

bzz:
After a brief discussion with @mcuadros today, we agreed that for now, in order to avoid GH rate-limit on bblfsh drivers on CI, instead of custom image that needs to be maintained, it would be better to:
 - mount some host dir to the bblfsh container under `/var/lib/bblfshd`
 - install only 1 driver i.e  `bblfshd bblfshctl driver install java` , instead of `--all/--recomended`
 - use [TravisCI cache](https://docs.travis-ci.com/user/caching/) to persist that dir

I.e one can see how [Gemini](https://github.com/src-d/gemini/blob/master/.travis.yml#L133) uses similar [approach](https://github.com/src-d/gemini/blob/master/docker-compose.yml#L40).

.travis.yml
```diff
@@ -31,6 +31,13 @@ jobs:
         - make lint
         - make coverage
         - make validate-commit
+    - script:
+      - docker run -d --name bblfshd --privileged carlosms/bblfsh-ci-test:v0.0.1
+      - mkdir $HOME/repos
+      - git clone https://github.com/src-d/gitbase-playground.git $HOME/repos/gitbase-playground
```
dpordomingo: `gitbase` [relies](https://github.com/src-d/gitbase/blob/master/Gopkg.toml#L22T) in `gopkg.in/src-d/go-git-fixtures.v3` for testing; I think we could also do it

carlosms: Maybe I'm missing the point, but what would that provide to our tests? It looks to me like a storage for fixture files, querying that git repository from gitbase-playground will not add any edge cases or anything like that.

dpordomingo: These fixtures seem to be the company standards as they can be seen in `go-git` and `gitbase` itself.
I'd also say that these fixtures could be more stable than our own repo, whose history could be rewritten at any point.

carlosms: So you are not proposing to clone the https://github.com/src-d/go-git-fixtures repo to `$HOME/repos`, you want to use the fixture contents to initialize the directory that the gitbase docker container will serve? I think that's a bit too complex and should be considered in a separate PR

dpordomingo: We're testing against [the own history of this repo](https://github.com/src-d/gitbase-playground/pull/41/files#diff-88a7e4c84767081f5b182aef7e4f9f35L33); I'm proposing to test against a standard fixtures repo: `go-git-fixtures`

About the implication of that change... I'm not sure about its size; I'd say it would be mostly to change the fixtures repo, and to change the `blob.hash` being tested from `fd30cea52792da5ece9156eea4022bdd87565633` to a new one: `82f2e6eef212f7a349a2e79cfd46e5118ea64f95`

About doing it in this or other PR (if approved) it'd be up to you ;)

.travis.yml
```diff
@@ -31,6 +31,15 @@ jobs:
         - make lint
         - make coverage
         - make validate-commit
+    - script:
+      - docker run -d --name bblfshd --privileged --volume $HOME/bblfsh-drivers/images:/var/lib/bblfshd bblfsh/bblfshd
```
smacker: You need to cache only images directory not all `/var/lib/bblfshd`. Because bblfsh writes there some other stuff that changes after every run. It will cause recreating and uploading cache every time.

carlosms: Thanks, It was a copy&paste mistake. The volume was supposed to be `$HOME/bblfsh-drivers`, the cache dir already points to `$HOME/bblfsh-drivers/images`.

.travis.yml
carlosms: As always, thanks for the suggestion. Since this PR is about adding the tests to Travis, I'd prefer to move this discussion to slack and merge this as soon as possible

carlosms:
Branch rebased and squashed! @smacker please take another look

smacker:
All good besides other jobs also download the same cache which isn't great.
It's easy to avoid with [matrices](https://docs.travis-ci.com/user/caching/#Caches-and-build-matrices), but I'm not sure how it works with stages. Maybe you can take a look?

If there is nothing in the documentation then it's not a big deal. According to logs setting the cache takes only 13 seconds. But it might be more problematic in the feature if we add more drivers.

carlosms:
I haven't seen an obvious way to set the cache for one stage only :thinking: 
Let's merge and keep improving the travis file as we do more testing.

PR #50, opened 2018-05-23
Sidebar

### LINKED ISSUE #19: Frontend: sidebar with reference
Add a left sidebar with the gitbase tables and columns reference in a collapsable structure.

This sidebar should also include a header with a small logo, and a footer with relevant links:
- https://sourced.tech
- https://github.com/src-d/gitbase
- https://doc.bblf.sh
- https://github.com/src-d/go-git

smacker:
Based on #38 
Fixes #19

No interaction with right side. We have another issue for it.

screen shot 2018-05-23 at 15 01 39

### DIFF
--- a/frontend/src/components/SampleQueries.js
+++ b/frontend/src/components/SampleQueries.js
@@ -1,0 +1,36 @@
+import React from 'react';
+import { Glyphicon } from 'react-bootstrap';
+import './SampleQueries.less';
+
+const queries = [
+  {
+    name: 'load all java files',
+    sql: 'select * from commits'
+  },
+  {
+    name: 'get uast from code',
+    sql: 'select * from commits'
+  },
+  {
+    name: 'top 50 repositories by something very long string',
+    sql: 'select * from commits'
+  }
+];
+
+function SampleQueries() {
+  return (
+    <div className="sample-queries">
+      <div className="title">Sample Queries</div>
+      <div className="list">
+        {queries.map((q, i) => (
+          <div key={i} className="query" title={q.name}>
+            <Glyphicon glyph="list" />
+            {q.name}
+          </div>
+        ))}
+      </div>
+    </div>
+  );
+}
+
+export default SampleQueries;
--- a/frontend/src/components/Sidebar.less
+++ b/frontend/src/components/Sidebar.less
@@ -1,0 +1,34 @@
+.sidebar {
+    display: flex;
+    flex-direction: column;
+
+    min-height: 100%;
+
+    .header {
+        flex: 0 0;
+
+        margin: 0 0 10px 0;
+        padding: 0 0 10px 0;
+
+        font-size: 16px;
+        font-weight: normal;
+
+        border-bottom: 1px solid #000;
+    }
+
+    .main {
+        flex: 1 1;
+        margin-bottom: 10px;
+    }
+
+    .footer {
+        flex: 0 0;
+
+        display: flex;
+        justify-content: space-between;
+    }
+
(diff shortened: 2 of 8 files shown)


frontend/src/components/SampleQueries.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { Glyphicon } from 'react-bootstrap';
+import './SampleQueries.less';
+
+const queries = [
+  {
+    name: 'load all java files',
```
dpordomingo: I'd avoid "todos" if possible:
- I'd rename first example as "get all commits", what matches this sql.
- And next examples I'd make them self explaining: removing them if they should be provided by business, or use simple ones like "get all repositories" and "get uast from code"

frontend/src/components/Sidebar.less
```diff
@@ -0,0 +1,34 @@
+.sidebar {
+    display: flex;
+    flex-direction: column;
+
+    min-height: 100%;
+
+    .header {
+        flex: 0 0;
+
```
dpordomingo: why do we need extra spaces everywhere in this file?

smacker: css properties are grouped and groups are split by a new line. That way it's easier to read. (at least for me)

dpordomingo: did we decide to do so in this project?
If we did so, we should add a linter to ensure the proper order and grouping for the shake of the consistency.

smacker: I don't know any linter rules which do that. I also don't know how to formalize it. We agreed to follow similar practices as we did in code-annotation. And we did it in code annotation: https://github.com/src-d/code-annotation/blob/master/src/components/Breadcrumbs.less

dpordomingo: I do not like to follow complex rules that are not programmatically enforced.
To do so, I think we could try to configure a linter like [stylelint.io](https://stylelint.io) + [stylelint-order](https://github.com/hudochenkov/stylelint-order), or analyze if we can use [ConcentricCSS](https://github.com/brigade/scss-lint/blob/master/data/property-sort-orders/concentric.txt) with less.

Otherwise, we're having inconsistencies as in https://github.com/src-d/code-annotation/blob/master/src/components/Experiment/Diff.less#L61

smacker: 1. I don't think any linter should be enforced for css blocks with <=3 rules or even 4-5 in some cases. Making it 6 lines instead of 3 would make it ugly and more difficult.
2. Using new line to split css rules to groups is quite common practice. For example:
- [Foundation UI](https://github.com/zurb/foundation-sites/blob/develop/scss/components/_switch.scss#L91)
- [Semantic UI](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/elements/button.less#L24)
as you could notice they also don't apply any grouping rules for classes with only a few lines
3. I thought a bit about how I split the groups. I have main groups like:
    1. layout mode: position/display/flex/...
    2. layout dimensions: width/height/margin/padding/...
    3. decoration: background/blox-shadow/...
    4. text: color/font-size/text-tranform/...
and use them when that makes sense. (usually when a group has more than 2 rules with layout mode as an exception)

If I didn't convince you, please be welcome to reformat it in your own style as soon as you start writing css for this project.

ricardobaeta:
Thanks @carlosms ! I'll get back to you on this pretty soon.

smacker:
yep, I also think we should improve it later when design is ready

Issue #43, opened 2018-05-23
Add queries examples to the sidebar


carlosms:
The user will have a few examples in the sidebar that, when clicked, will fill the input area.

Issue #22, opened 2018-05-09
Frontend: Perform 'describe table' queries from the sidebar


carlosms:
Clicking on a table name in the sidebar will automatically execute a 'describe table' query.

PR #51, opened 2018-05-23
handle clicks on schema and examples

smacker:
Based on #50

Fix: #22 
Fix: #43

Examples aren't real. We need a list of examples from the product team.

### DIFF
--- a/frontend/src/components/Schema.js
+++ b/frontend/src/components/Schema.js
@@ -17,14 +17,16 @@ class SchemaTable extends Component {
   }
 
   render() {
-    const { table, columns } = this.props;
+    const { table, columns, onTableClick } = this.props;
     const glyph = this.state.expanded ? 'minus' : 'plus';
 
     return (
       <div className="schema-table">
         <div className="name">
           <Glyphicon glyph={glyph} onClick={this.toggle} />
-          {table}
+          <span onClick={() => onTableClick && onTableClick(table)}>
+            {table}
+          </span>
         </div>
         {this.state.expanded && (
           <div className="columns">
@@ -47,17 +49,20 @@ SchemaTable.propTypes = {
       name: PropTypes.string.isRequired,
       type: PropTypes.string.isRequired
     })
-  ).isRequired
+  ).isRequired,
+  onTableClick: PropTypes.func
 };
 
-function Schema({ schema }) {
+function Schema({ schema, onTableClick }) {
   if (!schema) {
     return null;
   }
 
   return (
     <div className="schema">
-      {schema.map(item => <SchemaTable key={item.table} {...item} />)}
+      {schema.map(item => (
+        <SchemaTable key={item.table} {...item} onTableClick={onTableClick} />
+      ))}
     </div>
   );
 }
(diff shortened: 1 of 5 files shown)


frontend/src/components/SampleQueries.less
```diff
@@ -0,0 +1,21 @@
+.sample-queries {
+    .title {
+        padding: 3px 0;
+        margin-bottom: 10px;
+        background: #ccc;
+
```
dpordomingo: extra space

frontend/src/components/Schema.js
```diff
@@ -0,0 +1,80 @@
...
+import PropTypes from 'prop-types';
+import { Glyphicon } from 'react-bootstrap';
+import './Schema.less';
+
+class SchemaTable extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { expanded: false };
+
```
dpordomingo: spaces?

frontend/src/components/Sidebar.less
```diff
@@ -0,0 +1,34 @@
+.sidebar {
+    display: flex;
+    flex-direction: column;
+
+    min-height: 100%;
+
+    .header {
+        flex: 0 0;
+
```
dpordomingo: spaces?

frontend/src/components/SampleQueries.less
smacker: yes

frontend/src/components/Schema.js
smacker: yes

frontend/src/components/Sidebar.less
smacker: yes

frontend/src/components/Schema.js
dpordomingo: are we applying any special rule/logic in behind?
I saw other quite condense files, so I wonder about the difference in between...

smacker: I put an empty line between logic blocks:
```
super() - first block

this.state = {} - second

lots of bindings - third
```

smacker:
1. Hm. Current default SQL query doesn't work for me (gitbase schema is different, there is no repository_id field). Looks like the one you provided wouldn't also. I'm talking about this one:
```
SELECT COUNT(*) as num_commits, month, repo_id, committer_email FROM (
  SELECT MONTH(committer_when) as month, r.repository_id as repo_id, committer_email
  FROM repositories r
  INNER JOIN refs ON refs.repository_id = r.repository_id AND refs.ref_name = 'HEAD'
  INNER JOIN commits c ON YEAR(committer_when) = 2018 AND history_idx(refs.commit_hash, c.commit_hash) >= 0
) as t
GROUP BY committer_email, month, repo_id
```
And even if it will work with the next release of gitbase, I'm not sure that it's going to work in the release after next.

This PR implement frontend part of the sidebar interaction. I don't think invention of examples and catching up with gitbase which is in WIP state is in the scope of this PR.

2. 👍 (though it's weird that we don't have any handlers for them)

carlosms:
Yes, the examples I suggested work only on master, but in my opinion that's better than the ones that don't work on either on master or the latest release. If you don't want to commit to any concrete examples in this PR, what about removing the current ones and just use `/* To be done */` for all of them?

Issue #23, opened 2018-05-09
Frontend: Add code and UAST buttons in results table


carlosms:
Insert buttons in code and uast columns. For now the buttons will open a modal window with the raw contents.

PR #52, opened 2018-05-23
Add code and UAST buttons in results table

smacker:
Based on #38
Fix: #23

screen shot 2018-05-23 at 16 27 06
screen shot 2018-05-23 at 16 27 18
screen shot 2018-05-23 at 16 27 25

### DIFF
--- a/frontend/src/components/ResultsTable.js
+++ b/frontend/src/components/ResultsTable.js
@@ -1,11 +1,13 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import ReactTable from 'react-table';
+import { Button } from 'react-bootstrap';
 import 'react-table/react-table.css';
 import './ResultsTable.less';
 
 class ResultsTable extends Component {
   render() {
+    const { showCode, showUAST } = this.props;
     const columns = this.props.response.meta.headers.map(col => ({
       Header: col,
       id: col,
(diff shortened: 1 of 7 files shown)


frontend/src/components/ResultsTable.js
```diff
@@ -13,8 +15,15 @@ class ResultsTable extends Component {
         switch (typeof v) {
           case 'boolean':
             return v.toString();
+          case 'string':
+            // assume any multiline string is code
+            if (v.split('\n').length > 1) {
```
bzz: Would this in JS work for both platform-dependent line endings?

Would it be possible to include a simple test for this function, so further refactoring would rely on it?

On the side-note, in case >= 1 could work for this, may be some kind of equivalent of `.indeOf > 0` could be more performant..

bzz:
Also, here and in other pray rebase would make review life a bit easier.

frontend/src/components/ResultsTable.js
smacker: window line ending is `\r\n` so it will work.
👍 for `indexOf`.

smacker:
changed split to indexOf.

Added simple test. Though I don't see any value in it. Current logic is completely wrong (for example commit messages can be displayed as code according to it). And I don't think it's possible to solve on frontend. But we agreed to go with this simple-wrong way for now.

frontend/src/components/ResultsTable.js
```diff
@@ -14,8 +16,15 @@ class ResultsTable extends Component {
         switch (typeof v) {
           case 'boolean':
             return v.toString();
+          case 'string':
+            // assume any multiline string is code
+            if (v.indexOf('\n') > 1) {
```
carlosms: Should this be `> -1`? The way it is a multiline file with a blank line will not be detected.

smacker:
@bzz @carlosms please another round.

Issue #9, opened 2018-04-26
Modify uast nodes the same way as babelfish dashboard


carlosms:
From [this comment](https://github.com/src-d/gitbase-playground/pull/8#discussion_r184122947) by @smacker:
> dashboard replaces Role ID by Role name. Most probably playground should be compatible with it:
https://github.com/bblfsh/dashboard/blob/master/server/server.go#L217

PR #54, opened 2018-05-23
Modify uast nodes the same way as babelfish dashboard

smacker:
Fix: #9

### DIFF
--- a/server/handler/query.go
+++ b/server/handler/query.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/src-d/gitbase-playground/server/serializer"
 	"github.com/src-d/gitbase-playground/server/service"
-	"gopkg.in/bblfsh/sdk.v1/uast"
 
 	"github.com/go-sql-driver/mysql"
 )
@@ -155,26 +154,6 @@ func columnsInfo(rows *sql.Rows) ([]string, []string, error) {
 	}
 
 	return names, typesStr, nil
-}
-
-// unmarshallUAST tries to cast data as [][]byte and unmarshall uast nodes
-func unmarshallUAST(data interface{}) ([]*uast.Node, error) {
-	var protobufs [][]byte
-	if err := json.Unmarshal(*data.(*[]byte), &protobufs); err != nil {
-		return nil, err
-	}
-
-	nodes := make([]*uast.Node, len(protobufs))
-
-	for i, v := range protobufs {
-		node := uast.NewNode()
-		if err := node.Unmarshal(v); err != nil {
-			return nil, err
-		}
-		nodes[i] = node
-	}
-
-	return nodes, nil
 }
 
 // addLimit adds LIMIT to the query if it's a SELECT, avoiding '; limit'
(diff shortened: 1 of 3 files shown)


smacker:
As you may noticed in the source code it changes how roles are marshaled. Int -> String.

I don't think it should be handled by uast-viewer lib:
1. JS has no idea about role names, that info comes from proto file.
2. Theoretically it's possible for server to use newer version of uast with more roles than the lib was compiled with.
2. In my opinion it isn't responsability of the UI lib to convert input json. It should be agnotisc to input as much as possible. For example LA team wants to use that lib but with slighly different json. I think it's completely valid use case.

carlosms:
I understand your point of view. However, I still find it very odd that we want to have the server interfering with the json in this way, specially when it's a cosmetic change.

To have some context, and please correct me if I'm wrong, the only change this PR does is from
```json
  "Roles" : [
    18,
    1
  ],
```

to

```json
  "Roles" : [
    "Expression",
    "Identifier"
  ],
```

Even though this is a small change, I think it has big implications.
 
After this change our server is not returning UAST exactly as described in [the specification](https://doc.bblf.sh/uast/specification.html). This means that a hypothetical third party tool may work with the bblfshd output, but fail for the UAST we produce. Should we document the changes we make to the UAST? Will we edit the json for the CSV files in the same way? Should the uast-viewer lib document that you need to edit the UAST jsons before feeding them?

> JS has no idea about role names, that info comes from proto file.

The only purpose of the lib is to render UAST, so it is reasonable to expect that it knows about the role number-name relationship.

Let's say in the future we want to provide translations in different languages. You will need to have the list of role names and their translations (`Expression -> Expresión`), or use the number directly (`18 -> Expresión`).

> Theoretically it's possible for server to use newer version of uast with more roles than the lib was compiled with.

I think it's very reasonable to expect the uast-viewer lib to work with specific uast versions. If you rename the existing `children` property you wound't expect the uast-viewer to keep working. This is not a generic json renderer, the lib knows what a UAST is.

> In my opinion it isn't responsability of the UI lib to convert input json

I think this is the main point where we disagree. I see this change from number to string purely as a rendering task. The code is not converting the json to change its structure, it's just translating strings, which in my opinion should be responsibility of the front-end.

Having said all that, I understand this was probably already discussed for the bblfsh dashboard, and if @bzz agrees with @smacker's point of view I won't block the merge.

smacker:
Just some points to think about:
- The lib doesn't really depend on UAST format in the specification. It uses flat uast and there is a default transformer. You can feed any json if you are able to transform it to flat uast. And in long-term most probably it will be even less tightened to current uast. (for example it can render additional debug fields)
- The specification doesn't define JSON format for UAST. But protobuf. JSON format is an invention by dashboard to be able to send UAST to a browser. 
- The original scope of the lib was to extract UI component. Nothing else.
- As middle ground solution, we can export role names through endpoint and write another transformer in playground (though I don't really like that dashboard and gb-pg would work differently)

P.S. I don't really care about it. But just follow the original plan - reuse dashboard. If you don't like it, feel free to close this PR and create a new plan / issues.

smacker:
oh. I also forgot to mention you can't use this json to convert back to nodes in any language besides Go.
Because python would expect snake_case attributes and js cameCase with lowercase first letter.

bzz:
As discussed offline, it looks fine not to bring this complexity to the uast-viewer lib as:
 - bblfsh does not have standard JSON
 - bblfsh is going to move to uast v2 that is going to be flat, similar to what uast-viewer consumes right now

and thus keep it on the golang backend.

Issue #40, opened 2018-05-22
New endpoint to get the complete DB schema


carlosms:
Context: https://github.com/src-d/gitbase-playground/pull/35#discussion_r189810958

quoting @bzz:

> Instead for 
>  - API call `/tables`
>  - then, for each table call `DESCRIBE TABLE` from the frontend
> 
> We should have
>  - API call to `/tables`, so that the backend returns all this information in a single call
> 
> on top of per improvements (less HTTP calls), one of the purposes of having separate API endpoint on the backend was to abstract the details of how meta-information (schema) is collected.
> 
> Current approach, for a case if it is not a `DESCRIBE TABLE` - would require us to change both, backend and frontend. The one suggested above would requite only the backend change.

PR #55, opened 2018-05-23
tables endpoint should return complete schema

smacker:
Fix: #40

actually, the title of the issue and description are different. I implemented it according to the description.

### DIFF
--- a/server/handler/tables.go
+++ b/server/handler/tables.go
@@ -2,18 +2,58 @@ package handler
 
 import (
 	"net/http"
-	"strings"
 
 	"github.com/src-d/gitbase-playground/server/serializer"
 	"github.com/src-d/gitbase-playground/server/service"
 )
 
-// Tables returns a function that calls /query with the SQL "SHOW TABLES"
+// Tables returns DB schema
 func Tables(db service.SQLDB) RequestProcessFunc {
-	return func(r *http.Request) (*serializer.Response, error) {
-		req, _ := http.NewRequest("POST", "/query",
-			strings.NewReader(`{ "query": "SHOW TABLES" }`))
+	return func(r *http.Request) (res *serializer.Response, err error) {
+		rows, err := db.Query("SHOW TABLES")
+		if err != nil {
+			return nil, err
+		}
 
-		return Query(db)(req)
+		tables := make(map[string][]serializer.Column)
+
+		for rows.Next() {
+			var table string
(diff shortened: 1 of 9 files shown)


server/handler/tables.go
```diff
@@ -2,18 +2,58 @@ package handler
...
 import (
 	"net/http"
-	"strings"
 
 	"github.com/src-d/gitbase-playground/server/serializer"
 	"github.com/src-d/gitbase-playground/server/service"
 )
 
-// Tables returns a function that calls /query with the SQL "SHOW TABLES"
+// Tables returns DB schema
```
dpordomingo: If `Tables()` returns a `DB schema`, why not call the function `Schema` or something more meaningful?

server/handler/tables.go
```diff
@@ -2,18 +2,58 @@ package handler
...
 func Tables(db service.SQLDB) RequestProcessFunc {
-	return func(r *http.Request) (*serializer.Response, error) {
-		req, _ := http.NewRequest("POST", "/query",
-			strings.NewReader(`{ "query": "SHOW TABLES" }`))
+	return func(r *http.Request) (res *serializer.Response, err error) {
```
dpordomingo: I'd move this logic somewhere and I'd test it.
simple test cases, nothing complex:
- database with a couple of tables with a couple of cols,
- an empty database,
- an error,

server/handler/tables.go
smacker: PR is implemented according to the issue description. I think it's better to keep handler names and names of endpoints consistent.

server/handler/tables.go
smacker: we don't have such infrastructure for tests. And implementing it is definitely out of the scope of this PR.

server/handler/tables.go
smacker: yep. No problem with renaming if we want to change the endpoint name too!

carlosms: The initial design doc defined `/tables` to return only the tables, now that we have added more complete functionality I'm ok with renaming everything to `/schema`. @smacker go ahead if you feel it fits in this PR, otherwise we can consider this rename in a future PR.

carlosms:
Forgot to mention, `rest-api.md` should be updated.

smacker:
I renamed endpoint & updated docs.

Issue #42, opened 2018-05-23
Add export button


carlosms:
Instead of RUN the user will be able to click on EXPORT to download a file.
In this case the query will not automatically have the imposed `limit`.

The file format was not specified, will probably be csv.

PR #56, opened 2018-05-24
Add Export: backend and frontend

smacker:
Based on #55

Fix: #42 

screen shot 2018-05-24 at 15 39 25

The button opens new tab and endpoint streams csv to response.

### DIFF
--- a/server/handler/export.go
+++ b/server/handler/export.go
@@ -20,7 +20,7 @@ func Export(db service.SQLDB) http.HandlerFunc {
 			query := r.URL.Query().Get("query")
 			if query == "" {
 				return serializer.NewHTTPError(http.StatusBadRequest,
-					`Bad Request. Expected body: { "query": "SQL statement" }`)
+					`Bad Request. Query can't be empty.`)
 			}
 
 			rows, err := db.Query(query)
--- a/server/handler/export.go
+++ b/server/handler/export.go
@@ -1,0 +1,104 @@
+package handler
+
+import (
+	"database/sql"
+	"encoding/csv"
+	"net/http"
+	"strconv"
+
+	"github.com/go-sql-driver/mysql"
+	"github.com/src-d/gitbase-playground/server/serializer"
+	"github.com/src-d/gitbase-playground/server/service"
+)
+
+// Export returns a function that forwards an SQL query to gitbase and returns
+// the rows as CSV file
+func Export(db service.SQLDB) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		err := func(w http.ResponseWriter, r *http.Request) error {
+			query := r.URL.Query().Get("query")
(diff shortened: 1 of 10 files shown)


server/handler/export.go
```diff
@@ -0,0 +1,104 @@
...
+func Export(db service.SQLDB) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		err := func(w http.ResponseWriter, r *http.Request) error {
+			query := r.URL.Query().Get("query")
+			if query == "" {
+				return serializer.NewHTTPError(http.StatusBadRequest,
+					`Bad Request. Expected body: { "query": "SQL statement" }`)
+			}
+
+			rows, err := db.Query(query)
```
dpordomingo: `handler.Query()` and `handler.Export()` methods are almost the same... could not we abstract the common parts?

server/handler/export.go
```diff
@@ -0,0 +1,104 @@
...
+							b, err := sqlVal.Time.MarshalText()
+							if err != nil {
+								return err
+							}
+							record[i] = string(b)
+						}
+					case *sql.NullInt64:
+						sqlVal, _ := val.(*sql.NullInt64)
+						if sqlVal.Valid {
+							record[i] = strconv.FormatInt(sqlVal.Int64, 10)
```
dpordomingo: for example, why is this implementation different compared to [the other one in `handler.Query()`](https://github.com/src-d/gitbase-playground/blob/5967d42c9a4c206088d2be593ada537809a13dfc/server/handler/query.go#L111)?

server/handler/export.go
smacker: they are not that much the same. Most of the code of both handlers is switch block. But for `query` we return the original type (int, bool, ... in json) and for `export` we have to transform everything to string.
It's possible to "reuse" code. But then it would convert sql result to correct types -> correct types to string. 2 huge switch blocks instead of 1.

dpordomingo: fair enough

dpordomingo: fair enough

smacker: though we might create service for querying DB later. Then we can move this logic there.

server/handler/export.go
```diff
@@ -0,0 +1,104 @@
...
+						}
+					case *sql.NullString:
+						sqlVal, _ := val.(*sql.NullString)
+						if sqlVal.Valid {
+							record[i] = sqlVal.String
+						}
+					case *[]byte:
+						record[i] = string(*v)
```
carlosms: Why did you choose to not unmarshal UAST responses here?

carlosms:
Forgot to mention, `rest-api.md` should be updated.

smacker:
- Fixed proxy in dev env
- Updated docs
- Added transformation uast to json

PR #59, opened 2018-05-28
Add integration tests to coverage report

carlosms:
Based on #41
Part of #32

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ jobs:
       - git clone https://github.com/src-d/gitbase-playground.git $HOME/repos/gitbase-playground
       - docker run -d --name gitbase -p "3367:3306" -e "BBLFSH_ENDPOINT=bblfshd:9432" --volume $HOME/repos:/opt/repos --link bblfshd srcd/gitbase
       - sleep 15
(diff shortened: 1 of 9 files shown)


carlosms:
branch rebased on top of #41, now with docs updated

smacker:
I'm not sure counting code coverage by running integration tests is a good idea.

Mostly because you can't cover most of edge (error) cases with integration tests and need to use regular unit tests. Writing a successful path unit-test isn't such a big deal if you are writing fail path already.

Also, code coverage by integration tests may produce different results in code coverage in CI and on dev machine because they depend on the environment which can be annoying.

Third reason, integration tests require specific set up and if we use them as main tests instead of unit, it complicates developers experience.

carlosms:
Please note that the way it is running now the coverage will take into account **both** unit and integration tests. So this is not a proposal to replace unit tests and focus only on integration from now on, we should still implement more unit tests at some point.

But your comment has valid points, if others agree that integration tests should not be taken into account for the coverage I'm ok skipping this PR. cc @dpordomingo @bzz

bzz:
As discussed offline @carlosms is going to change where coverage is run

bzz:
@smacker I think all the feedback has been addressed and is ready for another pass

Issue #47, opened 2018-05-23
Code viewer


carlosms:
Integrate the uast-viewer lib to open code viewer when the code button (https://github.com/src-d/gitbase-playground/issues/23) is clicked.

Issue #61, opened 2018-06-05
Improve code detection and guess a language


smacker:
A language should be guessed using enry.

I propose to add information which column is a code and language in `meta` of the query endpoint response.

carlosms:
> I propose to add information which column is a code and language in meta of the query endpoint response.

Can you develop a bit more your idea? The current response `meta` contains information for the whole table, but the language is unique to each row.

e.g. in `select * from files` the column `blob_content` will not have a code button for binary files (like a `.png`), and each file may have a different language.

smacker:
The original idea was to figure out which column is a code and add it to a response and return array of languages for each row. Though now we decided to detect language using separate endpoint and call it only when a user opens a modal window. I'm not sure how to handle binary files.

smacker:
We do guess code & language now.
It's far from perfect but good enough for the current scope.
I'm closing the issue. Please create a new one if you see any particular cases that are needed to be improved.

PR #62, opened 2018-06-05
Code viewer

smacker:
Fix: #47

Based on #52, #60

This PR doesn't implement syntax highlighting in editor. It requires changes to code detection (move from FE to BE) and language guessing using envy. It will be implemented in another issue: #61

screen shot 2018-06-05 at 12 53 12

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import { Helmet } from 'react-helmet';
 import { Grid, Row, Col, Modal } from 'react-bootstrap';
 import SplitPane from 'react-split-pane';
-import UASTViewer, { transformer } from 'uast-viewer';
+import UASTViewer, { Editor, transformer } from 'uast-viewer';
 import 'uast-viewer/dist/default-theme.css';
 import Sidebar from './components/Sidebar';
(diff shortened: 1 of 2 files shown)


frontend/src/App.js
```diff
@@ -97,6 +108,33 @@ GROUP BY committer_email, month, repo_id`,
...
+      modalContent: <Editor code={code} />
+    });
+  }
+
+  showUAST(uast) {
+    this.setState({
+      showModal: true,
+      modalTitle: 'UAST',
+      modalContent:
+        // currently github returns only 1 item, UAST of the file
```
carlosms: You mean gitbase? :smile:

carlosms:
The only minor thing I see is how the modal overflows. As you can see in the screenshot, the vertical scroll is on the modal, while the horizontal is for the code component. If it makes sense, please consider changing this so both scrolls apply to the code component.

scrolls

frontend/src/App.js
smacker: oh. thanks! :-D

smacker:
Your suggestion makes perfect sense! The bottom scrollbar is kind of invisible on Mac so I missed it. Fixed.

PR #63, opened 2018-06-05
More unit tests

carlosms:
Part of #32.

Also contains a couple of fixes for bugs uncovered by the new tests.

### DIFF
--- a/server/handler/query.go
+++ b/server/handler/query.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/src-d/gitbase-playground/server/serializer"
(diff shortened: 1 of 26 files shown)


server/handler/schema_test.go
```diff
@@ -0,0 +1,65 @@
...
+func TestSchemaSuite(t *testing.T) {
+	suite.Run(t, new(SchemaSuite))
+}
+
+func (suite *SchemaSuite) TestGet() {
+	suite.mock.ExpectQuery("SHOW TABLES").WillReturnRows(
+		sqlmock.NewRows([]string{"table"}).
+			AddRow("foo"))
+
+	suite.mock.ExpectQuery("DESCRIBE TABLE.*").WillReturnRows(
```
smacker: I would expect `DESCRIBE TABLE foo` here to be sure the second query is correct.

server/handler/query.go
```diff
@@ -183,7 +184,10 @@ func addLimit(query string, limit int) string {
 		return query
 	}
 
-	query = strings.TrimRight(strings.TrimSpace(query), ";")
+	re := regexp.MustCompile(`\/\*(?s:.)*?\*\/`)
```
smacker: maybe better define & compile the regexp only once outside of the function?

carlosms:
Thanks @smacker. Since the feedback is not critical, I'll merge this PR as it is to fix the integration tests on Travis on master. I'll create a new PR shortly addressing the feedback.

PR #57, opened 2018-05-24
Parse endpoint

smacker:
Fix: #46

PR #64, opened 2018-06-06
Use debian instead of alpine for the docker container

carlosms:
The binary is dynamic since we added the bblfsh API client dependency, and it won't work in alpine anymore.

New Dockerfile based on the one we use in the [bblfsh dashboard](https://github.com/bblfsh/dashboard/blob/master/Dockerfile).


bzz:
👍 I belive the change that you mention is src-d/gitbase-playground#57

Issue #71, opened 2018-06-06
Add a history tab


carlosms:
The frontend will store (locally in the browser) a history of the queries.

More details about the design and functionality in the Visual Design & Pattern Library.
- query string
- elapsed time
- status (success/failure)
- failure reason
- start time
- We might want to store the tab title (editable)

Issue #78, opened 2018-06-06
Inform the user of the elapsed time


carlosms:
The frontend should count the time needed for each query, and reflect that in a success message

smacker:
Looks like we can't get time from gitbase (at least I don't see how).
But we can measure time on server or on frontend. What would you prefer?
@bzz @carlosms

carlosms:
I think the time as measured in the frontend should be fine.

PR #79, opened 2018-06-07
implement history tab

smacker:
fix: #71

screen shot 2018-06-07 at 07 55 54

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -8,6 +8,7 @@ import Sidebar from './components/Sidebar';
 import QueryBox from './components/QueryBox';
 import TabbedResults from './components/TabbedResults';
 import api from './api';
+import { STATUS_LOADING, STATUS_ERROR, STATUS_SUCCESS } from './state/query';
 import './App.less';
 
 class App extends Component {
@@ -27,6 +28,8 @@ FROM ( SELECT MONTH(committer_when) as month,
 ) as t GROUP BY committer_name, month, repo_id`,
       results: new Map(),
       schema: undefined,
+      history: [],
+
       // modal
       showModal: false,
       modalTitle: null,
@@ -51,7 +54,25 @@ FROM ( SELECT MONTH(committer_when) as month,
   }
 
   setResult(key, result) {
-    if (!this.state.results.has(key)) {
+    const { results, history } = this.state;
+    const historyIdx = history.findIndex(i => i.key === key);
+
+    if (historyIdx >= 0) {
+      const status =
+        typeof result.response !== 'undefined' ? STATUS_SUCCESS : STATUS_ERROR;
+      const newHistory = [
+        ...history.slice(0, historyIdx),
+        {
+          ...history[historyIdx],
+          status,
+          errorMsg: result.errorMsg
+        },
+        ...history.slice(historyIdx + 1)
+      ];
+      this.setState({ history: newHistory });
+    }
+
+    if (!results.has(key)) {
       // Tab was removed, ignore results
       return;
     }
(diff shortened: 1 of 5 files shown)


smacker:
the server doesn't return time now. So I left it out of the PR. We will need to add it here and under query box.

bzz:
Something @carlosms and I discussed yesterday was that it should be possible to have elapsed time counter for query on the frontend.

Question - would not it make sense to add time first, as #71 asks for it, and then rebase this work on that and include it in here?

smacker:
@bzz read the comment above. We have #78 for it.

frontend/src/components/HistoryTable.js
```diff
@@ -0,0 +1,64 @@
...
+    const { items, onOpenQuery } = this.props;
+    const columns = [
+      { Header: 'query', accessor: 'sql' },
+      {
+        Header: 'log',
+        id: 'log',
+        accessor: row => {
+          switch (row.status) {
+            case STATUS_SUCCESS:
+              return 'Query runned in Nsec';
```
carlosms: Query ran

frontend/src/App.js
```diff
@@ -49,7 +52,25 @@ FROM ( SELECT MONTH(committer_when) as month,
   }
 
   setResult(key, result) {
-    if (!this.state.results.has(key)) {
+    const { results, history } = this.state;
+    const historyIdx = history.findIndex(i => i.key === key);
```
carlosms: I think it might be cleaner (and faster) if we make `state.history` a Map, as `results`. What do you think?

frontend/src/components/HistoryTable.js
```diff
@@ -0,0 +1,64 @@
...
+      {
+        Header: 'date',
+        accessor: 'datetime',
+        Cell: ({ value }) => value.toString()
+      },
+      {
+        Header: 'action',
+        accessor: 'sql',
+        Cell: ({ value }) => (
+          <a onClick={() => onOpenQuery(value)}>open query</a>
```
carlosms: For some reason I don't get a pointer cursor on this link.

frontend/src/components/TabbedResults.js
```diff
@@ -116,6 +117,12 @@ class TabbedResults extends Component {
             </Tab>
           );
         })}
+        <Tab key="history" eventKey="history" title="History">
```
carlosms: According to the design this tab should float to the right

frontend/src/components/HistoryTable.js
smacker: copy past from design

frontend/src/App.js
smacker: it's not easy to serialize/deserialize map. In next PR I changed results too.

frontend/src/components/HistoryTable.js
smacker: That's correct. The link without `href` doesn't have pointer cursor. But before you were strongly against `href='#'`. Maybe I need to add addition css rule.

frontend/src/components/TabbedResults.js
smacker: I didn't move it to right on purpose. The design when we have too many tabs is undefined right now. It would affect implementation a lot.

smacker:
> History tab is hidden when you close all the results tab

the behavior for such case is undefined in design.

> The design has a 'reset history' link on top of the table

missed it. Thanks!

carlosms:
> the behavior for such case is undefined in design.

Still, I think it's reasonable to expect that the history tab should be visible when there are history entries, even if there are no result tabs open.

smacker:
1. rebased on master
2. added reset button
3. don't hide history when there are no other tabs
4. Corrected design typo

Issue #72, opened 2018-06-06
Edit tab titles


carlosms:
Each tab should have an edit button that will allow to edit the displayed tab title.

smacker:
should the title be replaced by input on this button click or it should be a modal dialog?

carlosms:
This will be better answered by @ricardobaeta.
I think it makes sense to replace the current title with an input. Pressing enter or clicking outside of the input would confirm the new name, and pressing escape would cancel the edit.

smacker:
@ricardobaeta please provide your input here. I also think it's better to replace the current title with an input but I would like to hear a confirmation from you.

ricardobaeta:
@carlosms @smacker It makes sense yes. You have my ok.

PR #80, opened 2018-06-07
Persist history and open tabs

smacker:
Based on #79

- results changed from Map to regular Object to be able to serialize and deserialize it easily. Also, it will be easier to migrate to redux later (if we do that)
- result keys are [nanoid](https://github.com/ai/nanoid) now instead of integers (it's like uuid but the library is super small). We need it to avoid any possible problems with collision during reading state from localStorage


bzz:
One thing, and sorry that it was not clear before - it is part of the product requirements NOT to persist anything, but a history. That includes open tabs. Rationale given by Product team was - it will be possible to recover all tabs from the history.

I do not think that is optimal, but as it was a requirement, so do you think we could follow that?

frontend/src/App.js
```diff
@@ -40,34 +68,76 @@ FROM ( SELECT MONTH(committer_when) as month,
...
+    if (historyIdx >= 0) {
+      const status =
+        typeof result.response !== 'undefined' ? STATUS_SUCCESS : STATUS_ERROR;
+      const newHistory = [
+        ...history.slice(0, historyIdx),
+        {
+          ...history[historyIdx],
+          status,
+          errorMsg: result.errorMsg
+        },
```
bzz: Could you please help me understand, if a tab title gets persisted in the history?

frontend/src/App.js
```diff
@@ -40,34 +68,76 @@ FROM ( SELECT MONTH(committer_when) as month,
...
 
-    this.setState({ results: loadingResults });
+    this.setState({
+      results: loadingResults,
+      history: [
+        {
+          key,
+          sql,
+          datetime: new Date(),
+          status: STATUS_LOADING
```
bzz: Could you please help me understand, if a tab title and all the other fields from https://github.com/src-d/gitbase-playground/issues/71 get persisted in the history entities?

smacker:
@bzz when the requirement has changed? It was the opposite - keep open tabs (and use suspend feature) during all previous discussions I participated in.

frontend/src/App.js
smacker: a current title is sql and as you can see I persist it. We will have a separate field for the title in #72

bzz:
> when the requirement has changed? It was the opposite - keep open tabs (and use suspend feature) during all previous discussions I participated in.

I do not think they have changed, but they might have been communicated poorly :/
Sorry, I know, this can be frustrating.

As mentioned above: 
 - I agree, it makes to persist the tabs, but..
 - but we were explicitly requested by @mcuadros to persist only 1 piece of state - the history tab (with rationale mentioned above)

If that is something you personally strongly disagree with - we can try escalate this and discuss it again, esp in context of research that @ricardobaeta did [recently](https://docs.google.com/document/d/1Q2P2xYnLf8cjer2wRlP6asiKky_lNd1rObrWj1Bmir0/edit#heading=h.pgd0owi6ohb7) and possible changes that it might bring.

Meanwhile, to move forward with this I would suggest following the request.

smacker:
@bzz it's super weird. I definitely remember discussing saving open tabs and Maximo was there. Here is no need to hurry with this PR. Let's discuss it with product team before moving forward to avoid wasting the time of removing and reimplementing the feature again.

mcuadros:
@marnovo if everyone finds weird this, lets study this and design the UX for this.

marnovo:
@smacker @bzz thanks for bringing this to our attention!
Will discuss with @mcuadros and @ricardobaeta this week in light of the UX research results.

bzz:
Sounds good, thank you for update. As discussed today, let's keep this PR open until this is designed by Product team.

ricardobaeta:
@bzz @smacker @marnovo @mcuadros  

Thank you all for the time and effort you put in articulating on this, and sorry for jumping onboard just now.

If I'm understanding what's being discussed here, the issue to address relies upon:

#### User Journey

*A user goes to the playground, plays with it for a while and close the browser tab. Later on, he remembers about this awesome product and goes again to the playground, entering the URL in the browser.*

#### What UX behaviour to expect

**1)** He gets **exactly** what he had in the previous session

- The last typed query on the query input 
- The corresponding opened tab with the results
- All the other tabs he had
- The History tab with the queries history
- Saved queries [1](https://docs.google.com/document/d/1Q2P2xYnLf8cjer2wRlP6asiKky_lNd1rObrWj1Bmir0/edit#bookmark=id.gwk2u75mgqmt) & [2](https://docs.google.com/document/d/1Q2P2xYnLf8cjer2wRlP6asiKky_lNd1rObrWj1Bmir0/edit#bookmark=id.9yxtti7tp5ck)
-  Statistics [1](https://docs.google.com/document/d/1Q2P2xYnLf8cjer2wRlP6asiKky_lNd1rObrWj1Bmir0/edit#bookmark=id.jiybglzbuxll)

> he should be logged in

**2)** He gets the default new session with **some perks**

- The product's default initial state (default query & correspondent result)
- The History tab with the queries history, stored on the browser from his previous session

- - -

My take on this is that, until the time comes for us to cover power user needs, some kind of "my playground" as explained on **behaviour 1)**, we should stick with what we aimed to accomplish, ditto **behaviour 2)**, for at least two reasons:

1) There was no research findings about having persisting tabs
2) Stakeholders had shared vision upon *session perks*  

I totally believe on improving whilst building, and therefore I appreciated a lot your input Maxim, nonetheless I take responsibility for not making this clear earlier.

smacker:
@bzz @carlosms please take a look

frontend/src/App.js
```diff
@@ -14,6 +15,28 @@ import './App.less';
...
 const INACTIVE_TIMEOUT = 3600000;
 
+function persist(key, value) {
+  window.localStorage.setItem(key, JSON.stringify(value));
+}
+
+const dateFormat = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/;
+
+function jsonReviver(key, value) {
+  if (typeof value === 'string' && dateFormat.test(value)) {
```
carlosms: Since we are only storing history elements I think we could simplify the regexp to 
`if (key === 'datetime')`

smacker: In my opinion, it's better to have a common solution for datatime fields than relying on the particular structure of currently saved data.

PR #81, opened 2018-06-07
Make the sidebar collapsable

smacker:
Fix: #77

screen shot 2018-06-07 at 19 07 18
screen shot 2018-06-07 at 19 07 25


carlosms:
I've tested it and the main body does not resize to take up the new free space left by the collapsed sidebar, which I believe should be the expected behaviour.

smacker:
You are right. Repushed. I removed cols completely because even 1 col is way too much for the collapsed sidebar.

PR #82, opened 2018-06-08
Add .drone.yml file

### LINKED ISSUE #74: Create a drone.yaml file

carlosms:
Fixes #74 

The file is based on the [one in code-annotation](https://github.com/src-d/code-annotation/blob/master/.drone.yml).

With the new configuration drone will create, push, and deploy to staging a new docker image for each commit in master. It will also create and push when a new tag is created.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -29,8 +29,8 @@ docker_image: &docker_image
 helm_deploy: &helm_deploy
   image: quay.io/ipedrazas/drone-helm:master-9b37211
   skip_tls_verify: true
-  helm_repos: charts=https://github.com/src-d/charts.git
-  chart: charts/gitbase-playground
+  helm_repos: srcd-charts=https://src-d.github.io/charts/
+  chart: srcd-charts/gitbase-playground
   release: gitbase-playground
   tiller_ns: kube-system
   wait: true
--- a/.drone.yml
+++ b/.drone.yml
@@ -9,13 +9,10 @@ branches: [master]
 build: &build
   image: golang:1.10-stretch
   commands:
-    - curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
-    - . $HOME/.nvm/nvm.sh
-    - nvm install 8
-    - nvm use 8
-    - npm install -g yarn
+    - curl -sL https://deb.nodesource.com/setup_8.x | bash
     - apt-get update
-    - apt-get install -y --no-install-recommends --no-install-suggests libxml2-dev 
+    - apt-get install -y --no-install-recommends --no-install-suggests nodejs libxml2-dev
+    - npm install -g yarn
     - make dependencies
     - PKG_OS=linux DOCKER_OS=linux DOCKER_ARCH=amd64 make packages
   debug: true
@@ -66,7 +63,8 @@ pipeline:
     <<: *helm_deploy
     prefix: STG
     secrets: [ STG_API_SERVER, STG_KUBERNETES_TOKEN ]
-    values: gitbasePlayground.image.tag=commit-${DRONE_COMMIT_SHA:0:7},gitbaseServer.image.tag=v0.12.1-rc1,ingress.globalStaticIpName=gitbase-playground-staging,ingress.hostname=gitbase-playground-staging.srcd.run
+    values: gitbasePlayground.image.tag=commit-${DRONE_COMMIT_SHA:0:7},ingress.globalStaticIpName=gitbase-playground-staging,ingress.hostname=gitbase-playground-staging.srcd.run
+    values_files: [.helm-staging.yml]
     when:
       branch: [master]
       event: [push]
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -0,0 +1,87 @@
...
+  path: src/github.com/src-d/gitbase-playground
+
+branches: [master]
+
+# ANCHORS
+
+build: &build
+  image: golang:1.10-stretch
+  commands:
+    - curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
```
smacker: is there any specific reason to install nvm instead of nodejs directly?
```
curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
sudo apt-get install -y nodejs
```

.travis.yml
```diff
@@ -65,11 +65,6 @@ jobs:
         skip_cleanup: true
         on:
           all_branches: true
-    # push to dockerhub
```
smacker: any specific reason to move it to drone?

bzz:
@carlosms please help me understand, what would be the best way to fix default bblfsh version? 

And then what would be the best way to pass versions of both bblfsh and `gitbaseServer.image.tag` so we could control those on our side?

.travis.yml
rporres: Yes, for the moment we're not doing CD with Travis, hence the move

rporres: Yes, if we kept build in Travis, drone would have to wait for the image to be built. I don't like that dependency, I see no good things coming from that strategy

rporres:
Being the app not straight forward, there are a number of things that can be configured

My golden rule is to avoid deploys using `latest` tags. It's not a good idea.

* In bblfsh server you can fix server and driver versions
* In gitbase server you can fix server version and the database repositories. In order to be consistent, I would fix the repositories to a specific commit so that you always have the same version (the chart pulls `HEAD` otherwise everytime it gets deployed)

There are a number of things to control. I would consider using a value files if you don't like maintaining a big `values` key. It doesn't look that it interpolates things like `${DRONE_TAG}`, so you may have to keep also `values` and let `helm` command merge both

bzz:
Sounds very good, thank you for explanation @rporres ! 

@carlosms Let's fix make sure we fix versions of both, gitbase and bblfsh. 

For simplicity and because it just an internal staging and not a web service that needs to provided/maintain - I would say passing vars pre-configured in Drone for those images should be fine (to avoid image building on commits/tags \w file changes)

.drone.yml
carlosms: I used nvm to copy the exact same steps we have in `.travis.yml`, but now it's changed to your suggestion in a new commit.

carlosms:
@rporres  I've pushed a new commit to address your comments, please take a look.

I didn't set a fixed commit for each repo for now. The reason is that I'm not so sure that we can really replicate the exact same DB results with a `git clone` + `git reset`.
If I set a specific commit, a new deployment a week from now will still return new tags  for `select * from refs`.

.helm-staging.yml
```diff
@@ -0,0 +1,49 @@
...
+    gitbase:
+      url: https://github.com/src-d/gitbase.git
+    gitbase-playground:
+      url: https://github.com/src-d/gitbase-playground.git
+    bblfshd:
+      url: https://github.com/bblfsh/bblfshd.git
+  image:
+    repository: srcd/gitbase
+    tag: v0.12.1-rc1
+    pullPolicy: IfNotPresent
```
rporres: Why are you overwriting here with values that are exactly the same as the default values set in chart's `values.yaml` in a lot of cases? I'm thinking about image repositories, for example

.drone.yml
```diff
@@ -0,0 +1,85 @@
...
+  registry: docker.io
+  repo: srcd/gitbase-playground
+  secrets: [ docker_username, docker_password ]
+  dockerfile: Dockerfile
+  debug: true
+
+helm_deploy: &helm_deploy
+  image: quay.io/ipedrazas/drone-helm:master-9b37211
+  skip_tls_verify: true
+  helm_repos: charts=https://github.com/src-d/charts.git
```
rporres: I don't think a git repository is what `drone-helm` is expecting, judging from https://github.com/ipedrazas/drone-helm#using-private-repositories documentation

I'd test with something like

```
helm_repos: srcd-charts=https://src-d.github.io/charts/
chart: srcd-charts/gitbase-playground
```
also setting a different name to avoid using `charts`, that seems quite generic

.helm-staging.yml
carlosms: Since the default values are in another repository, I wanted to leave here the common changes we might want to make, for visibility.

.drone.yml
carlosms: thanks, fixed in 81194b3.

.helm-staging.yml
rporres: Hmmm, I don't think you're going to change that, but it's fine by me :)

PR #83, opened 2018-06-12
Inform the user of the elapsed time

### LINKED ISSUE #78: Inform the user of the elapsed time

smacker:
Fix #78 

Based on #79 

I'm super confused that results and the box with the time are in different places. It looks like they are disconnected, so I implemented them like that. Box shows information about the last result. But I'm not sure it's correct.

screen shot 2018-06-12 at 12 58 13


carlosms:
> I'm super confused that results and the box with the time are in different places. It looks like they are disconnected

My understanding of the design spec is the same, the result message is not attached to the current active tab, but to the last executed query.

PR #85, opened 2018-06-12
Detect lang endpoint

smacker:
As discussed we need such endpoint. There is a problem that enry can't detect language without filename in most cases. So I'm not sure we need to merge it now. But in case we need.


bzz:
> As discussed we need such endpoint

Just for the context, we need it for 2 reasons
 - Code -> UAST, for request to bblfsh
 - Syntax hightligh for CodeMirror

bzz:
@smacker how about handling the case of detecting a language only by content as well?

smacker:
@bzz you can see in tests you can pass empty filename and enry will try to guess it by content only https://github.com/src-d/gitbase-playground/pull/85/files#diff-bbb690ba3ea1b1821225ad2e890527c9R35

smacker:
There is no any custom logic. It's just HTTP proxy to enry.
Documentation of enry doesn't mention it as a special use-case: https://godoc.org/gopkg.in/src-d/enry.v1#GetLanguage

Issue #73, opened 2018-06-06
Suspend inactive tabs


carlosms:
The basic requirement is to have tabs drop the loaded information and go into a suspended state after a specific period of time being hidden.
The suspended tab will have a "reload" button to run the query again.

PR #87, opened 2018-06-13
Suspend inactive tabs

smacker:
Fix: #73

screen shot 2018-06-13 at 14 07 53

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -11,6 +11,8 @@ import api from './api';
 import { STATUS_LOADING, STATUS_ERROR, STATUS_SUCCESS } from './state/query';
 import './App.less';
 
+const INACTIVE_TIMEOUT = 60000;
+
 class App extends Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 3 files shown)


frontend/src/App.js
```diff
@@ -11,6 +11,8 @@ import api from './api';
 import { STATUS_LOADING, STATUS_ERROR, STATUS_SUCCESS } from './state/query';
 import './App.less';
 
+const INACTIVE_TIMEOUT = 60000;
```
bzz: Let's set the default to 1h

carlosms:
I wonder if we should make `state.results` store `./state/query` constants, adding a new `STATUS_SUSPENDED`.

This way we avoid the current logic in `TabbedResults.js render()`, asking for `query.loading`, `query.errorMsg`, `query.response`.

smacker:
@carlosms currently we use constants only in history. And history doesn't have suspended state.
But in general, I would like to refactor it that way. In separate PR though. (after styling maybe?)

Issue #44, opened 2018-05-23
Apply product team design


carlosms:
Apply the design details, as provided by the product team

bzz:
The design, as well as means of access to it, is provided under https://github.com/src-d/product/issues/264#issuecomment-392651630

carlosms:
Closing as most of the work is done. Further changes and fixes can be done in new small issues.

PR #95, opened 2018-06-14
Set styling for results table

carlosms:
Part of #44.

a

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -1,10 +1,10 @@
 import React, { Component } from 'react';
 import { Helmet } from 'react-helmet';
-import { Grid, Row, Modal, Button } from 'react-bootstrap';
-import { bootstrapUtils } from 'react-bootstrap/lib/utils';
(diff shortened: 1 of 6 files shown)


carlosms:
> 1. CI is failing

Fixed, test snapshot needed to be updated.

> 2. On the design I don't see a shadow from the columns header

I left it from the original React Table style because otherwise the header and the rows kind of mixed. Then I realized I forgot to set actual borders from the design. Fixed now.

> 3. Vertical line between pages & num of rows is missed
> 4. Margins for next/prev buttons are incorrect (on the design they are located much closer to center)

I've made a few changes to make it look closer to the design:

a

smacker:
great job. Only one suggestion:
Move `bootstrapUtils.addStyle`s to separate file and include them in index.js/setupTest.js instead of copy-pasting.

carlosms:
Thanks, that makes sense. I've moved it to a new `utils` dir. Merging!

PR #100, opened 2018-06-15
Fix query and edit button misalignment

carlosms:
Part of #44.

The EDIT button inside the query results is misaligned with the SQL query, because they use different font family and size. With a fixed line-height both fonts align at the baseline.

a

### DIFF
--- a/frontend/src/components/TabbedResults.less
+++ b/frontend/src/components/TabbedResults.less
@@ -32,6 +32,10 @@
     .query-row {
         margin-bottom: 2em;
 
+        .query-text, .edit-query {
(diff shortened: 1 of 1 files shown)


carlosms:
This PR fixes the alignment regardless of the number of lines, truncating the lines is an unrelated change waiting on feedback on the design part.

ricardobaeta:
@smacker 

> As I remember only 1 line of the query should be shown. @ricardobaeta am I right?

Yes, with ellipsis :) Thank you!

carlosms:
Changes merged in #122

Issue #97, opened 2018-06-15
Wrong behaviour tab title edits


carlosms:
- When the pencil icon is clicked the new textarea does not get focused automatically. This is confusing because you start writing and nothing happens
- If you click on more than one pencil icons, you can have a few tab titles as textarea. I think this could be related to the previous point, if the textarea does not get focused the event on focus lost does not get fired.
- There space key does not work
- The arrow keys (or ctrl+arrow) do not move the cursor, they change between tabs

smacker:
everything was fixed

PR #101, opened 2018-06-15
Add autofocus to tab title input

smacker:
Fix first 2 points in #97

The "space" problem comes from here: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/SafeAnchor.js#L61
Arrows problem comes from here: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Nav.js#L238

We can fix it by inheriting bootstrap components and disable handlers when user clicked "edit".
Please let me know if you are ok with such solution. (I'll implement in separate PR)

### DIFF
--- a/frontend/src/components/DivTabs.js
+++ b/frontend/src/components/DivTabs.js
@@ -1,0 +1,155 @@
+/* eslint-disable */
+
+/*
+Copy-past from react-bootstrap
+We can't just extend original component and redefine renderTab method because it exports uncontrollable HOC.
+
+We need to use <div> instead of <a> as tab titles due to Firefox behaviour of <input> inside <a>.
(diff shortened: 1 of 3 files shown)


carlosms:
About:
> We can fix it by inheriting bootstrap components and disable handlers when use clicked "edit".
> Please let me know if you are ok with such solution. (I'll implement in separate PR)

I was hoping we could simply stop the event in the `<input>`, something like
```
            onKeyDown={e => {
              e.stopPropagation();
            }}
```
Which testing quickly works for me. Any caveats with this options?

smacker:
Fixed firefox. Though a bit dirty. But I couldn't come up with better solution.

I usually try to avoid using `stopPropagation` it makes things very hard to debug later. But in this particular case, it's alright in my opinion. Thanks a lot for suggestion!

PR #109, opened 2018-06-18
Use svg icons

carlosms:
Part of #44.

Uses svg icons as inline `<svg>` elements thanks to [svg-react-loader](https://github.com/jhamlet/svg-react-loader).

Missing icons from the design spec:
- Error message. The error message needs to be styled and moved to the same place as the success message, it makes more sense to add the icon in that PR.
- Main big icons for "running query" and "suspended tab": Those svg files will be provided separately soon.

**Important**: before merging we need confirmation from @ricardobaeta if we need to add any new license notice for the icons.

a


ricardobaeta:
@carlosms The Icons I've used - and changed - are from this designer and are free. No need to refere a license.

https://dribbble.com/shots/2888226-1800-Free-Minimal-Icon-Pack-20x20

There's just this note:

> Note: Please don’t freeboot this pack on your personal website of resources without noticing me. Thank you!

Should we contact the Designer?

carlosms:
> > Note: Please don’t freeboot this pack on your personal website of resources without noticing me. Thank you!
> 
> Should we contact the Designer?

The icons do not provide any kind of license, I don't know if this can be problematic. I kindly request @marnovo to please take a look.

frontend/src/icons/close-query-tab.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
```
smacker: let's run [svgo](https://github.com/svg/svgo) on the icons. You convert them to components, but I think will still help to make them more optimal.

frontend/src/components/SampleQueries.less
```diff
@@ -22,5 +22,17 @@
         text-overflow: ellipsis;
         cursor: pointer;
         padding-right: 40px;
+
+        svg path {
```
smacker: I see you repeat this rules multiple times. Do you think we can create function `icon-variant` to use here? And reuse it in `icon-btn-variant`?

frontend/src/icons/close-query-tab.svg
carlosms: Before modifying these icons, let's confirm with @ricardobaeta's it's safe.

Are these files the sources you would use to edit, or just exported versions? in other words, if we modify these files, would that prevent you from going into this repository to tweak some icons?

ricardobaeta: Hi Carlos, as I understood the icons are Free, you just can't put it in your website - the originals -  as a resource without noticing the designer. We used the icons and we've changed every one of them. They will be in this repo yes, but we're not redistributing the designer's work. Do you agree? @marnovo What's your take on this?

ricardobaeta:
@carlosms  @bzz 

> Main big icons for "running query" and "suspended tab": Those svg files will be provided separately soon.

Here you have the gif animations for the "running query" & "suspended tab".

Drive Folder: https://drive.google.com/drive/u/1/folders/1FiUHRycMnQJnu8MZz-EQSjfxjoBZqSMj

alex-jumping

alex-suspended

frontend/src/icons/close-query-tab.svg
carlosms: In this thread what I'm asking is not really related to the license. What Maxim is suggesting is to modify the `.svg` files here to make them simpler and smaller (internal structure only, without changing the resulting image). And what I'm asking you to confirm is whether or not you would care if we do something to these files that prevents you from opening them, and working in this repository files directly.

I assume you have your files somewhere else and these are just exports that you don't plan to open and modify directly, but just to double check. Thanks!

frontend/src/components/SampleQueries.less
carlosms: I did a small refactor in 1de485b955aea153f3ca200ec272c70e077f2234.

carlosms:
The animations look great :smile: 

They are already added to this branch.

a

b

carlosms:
Rebased on top of master to fix conflicts

carlosms:
Rebased again.

PR #114, opened 2018-06-19
Fix history tab hover color

carlosms:
Fix a bug introduced in #112, repeated `.history` class name.


carlosms:
Already fixed in #116.

PR #118, opened 2018-06-20
Set styles for uast-viewer & modals

carlosms:
Part of #44.

Sets the overall modal styles, and the fonts and colors for the code and uast viewers.
The only difference with the design spec is the code viewer font, set to monospace instead of 'source sans'.

a
b

### DIFF
--- a/frontend/src/App.less
+++ b/frontend/src/App.less
@@ -152,7 +152,7 @@ body {
     }
 
     .uast-title-collapsible.collapsed {
-        background-color: rgba(27, 120, 183, 0.04); // @primary, 0.04 opacity
+        background-color: fade(@primary, 4%);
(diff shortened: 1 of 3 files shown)


frontend/src/App.less
```diff
@@ -100,7 +106,65 @@ body {
...
+    .uast-property-name {
+        color: @uast-black;
+    }
+
+    .uast-title-collapsible:hover > span {
+        border-bottom: 1px dashed @uast-black;
+    }
+
+    .uast-title-collapsible.collapsed {
+        background-color: rgba(27, 120, 183, 0.04); // @primary, 0.04 opacity
```
smacker: maybe we can use `fade` function here?

Issue #127, opened 2018-06-21
Add x-path query and settings panel to uast viewer pane


smacker:
Only for the right side of `Code` modal for now.

- Add show location checkbox
- Allow user to change bblfsh server
- Add x-path query panel
- Handle results of the search correctly

PR #128, opened 2018-06-21
Add x-path query and settings panel to uast viewer pane

smacker:
Fix: #127

screen shot 2018-06-21 at 16 04 39

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -126,12 +126,14 @@ function detectLang(content, filename) {
   }).then(res => res.data);
 }
 
-function parseCode(language, content) {
+function parseCode(language, content, filter, customServerUrl) {
   return apiCall('/parse', {
     method: 'POST',
     body: {
       language,
-      content
+      content,
+      filter,
+      serverUrl: customServerUrl
     }
   }).then(res => {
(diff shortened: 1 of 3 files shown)


smacker:
I think it's product decision, not my. Do we want filtering on code & uast modals now or do we want wait?
Uast filtering on another modal depends on this PR.

smacker:
Rebased on master. Rebase was hard and commit history didn't make sense anymore. So I squshed commits.

bzz:
👍  \cc @carlosms for a review

carlosms:
I see the following problem. The `CodeViewer.state.filter` acts as a global setting, persistent between different modals (different rows). But visually the checkbox does not follow this setting.

To reproduce:
- open a CODE viewer for one row
- set a custom bblfsh url
- close this modal
- open modal for other row

Now the new modal will use the bblfsh url set previously, but the "custom server" checkbox is unchecked.

I would also use  `127.0.0.1:9432` instead of `0.0.0.0:9432`.

smacker:
thanks for reporting. side effect after rebase. I'll fix it.

smacker:
Fixed.

I use `0.0.0.0:9432` because dashboard uses it.

PR #132, opened 2018-06-22
Show error messages in main query box

carlosms:
Part of #44.

Implements the design spec for error messages.

Now when an error happens:
- The error message is displayed in the main query box
- The loading tab is deleted

Other change: the last success or error message is cleared when the 'run query' button is clicked.

a

### DIFF
--- a/frontend/src/components/QueryBox.js
+++ b/frontend/src/components/QueryBox.js
@@ -13,6 +13,36 @@ import 'codemirror/addon/hint/sql-hint';
 
 import './QueryBox.less';
 import SuccessIcon from '../icons/success-query.svg';
+import ErrorIcon from '../icons/error-query.svg';
+
+function ResultInfo({ result }) {
+  if (!result) {
+    return null;
+  }
+
+  if (result.response && result.response.meta) {
+    return (
+      <span className="meta meta-success">
+        <SuccessIcon className="big-icon" />Showing rows (query took{' '}
+        {result.response.meta.elapsedTime / 1000} seconds)
+      </span>
+    );
+  }
+
+  if (result.errorMsg) {
+    return (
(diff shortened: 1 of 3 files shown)


frontend/src/components/QueryBox.js
```diff
@@ -70,11 +71,17 @@ class QueryBox extends Component {
...
+    if (result && result.response && result.response.meta) {
       meta = (
-        <span className="meta">
+        <span className="meta meta-success">
           <SuccessIcon className="big-icon" />Showing rows (query took{' '}
-          {resultMeta.elapsedTime / 1000} seconds)
+          {result.response.meta.elapsedTime / 1000} seconds)
+        </span>
+      );
+    } else if (result && result.errorMsg) {
```
smacker: could you please move it `meta` to functional component?
it would simplify code. something like:
```
if (!result) {
  return null;
}
if (result.errorMsg) {
  return ...
}
...
```

carlosms: Done in e549534

frontend/src/components/QueryBox.js
```diff
@@ -13,6 +13,34 @@ import 'codemirror/addon/hint/sql-hint';
...
+  }
+
+  if (result.response && result.response.meta) {
+    return (
+      <span className="meta meta-success">
+        <SuccessIcon className="big-icon" />Showing rows (query took{' '}
+        {result.response.meta.elapsedTime / 1000} seconds)
+      </span>
+    );
+  } else if (result.errorMsg) {
```
smacker: you don't need else here.

carlosms: Changed in 1c7927b

Issue #120, opened 2018-06-20
SQL input box is broken in FF


smacker:
1. If the height of the box isn't too big I can't scroll down to see the last line of the default query
2. Selection works weird:
<img width="753" alt="screen shot 2018-06-20 at 17 39 00" src="https://user-images.githubusercontent.com/406916/41669057-24bb6230-74b1-11e8-89e3-c7d65f334346.png">

carlosms:
More info: works fine for chrome and firefox 60.0.2 on linux.

PR #134, opened 2018-06-22
Reload codemirror when fonts are loaded

smacker:
Fix: #120

Apparently, it is broken not only in FF but in Chrome too.


smacker:
you could try to slow down your network and load it without cache. The bug happens when JS kicks before the font is downloaded.

Issue #49, opened 2018-05-23
UAST filtering from the frontend


carlosms:
To allow UAST XPATH queries in the UAST viewer we would need a new endpoint, `POST /filter`, to ask babelfish to apply an XPATH query to a UAST.

This endpoint will need the protobuf UAST data, which we currently do not send to the frontend.
We will need to add the protobuf data to the `POST /query` response; and decide if we want to decode the protobuf data in the frontend, or send both json and protobuf.

PR #135, opened 2018-06-22
Uast filtering

smacker:
Based on: #128

Fix: #49

No protobuf decoding on FE. If we decided to go that way: we can implement it later.
Though I would experiment on dashboard first, not here.

screen shot 2018-06-22 at 15 56 24


bzz:
@smacker shall we add simple progress indicator in here? Nothing fancy, a simple clue to the user that it's in progress.

On my local machine under heavy CPU load it took a while to UAST to get back and having no visual feedback in situation like this can be confusing

bzz:
On interaction between code-uast: do you think it could be possible that we have off-by-one error when choosing a character under cursor?

screen shot 2018-06-26 at 10 10 01 am

Also, I have noticed that when clicking on the UAST it only moves cursor to the position, but not the scroll of the code - sometimes it is quite had to follow esp on the filtered UAST.

bzz:
Few other things that seems not to be related to the UAST filtering itself were moved to https://github.com/src-d/gitbase-playground/issues/137

smacker:
Yes. You are right. I missed a loader somehow, my bad. Most probably it worked too fast on my laptop :) Thanks!

frontend/src/components/UASTViewerPane.js
```diff
@@ -0,0 +1,126 @@
...
+  handleCustomServerChange,
+  handleFilterChange,
+  handleSearch
+}) {
+  const searchResults = getSearchResults(uastViewerProps.uast);
+  const rootIds = searchResults || [ROOT_ID];
+
+  let content = null;
+  if (searchResults && !searchResults.length) {
+    content = <NotFound>Nothing found</NotFound>;
```
carlosms: should this be `content = <NotFound />;`?

server/handler/uast.go
```diff
@@ -79,9 +80,52 @@ func Parse(bbblfshServerURL string) RequestProcessFunc {
...
+		}
+
+		nodes, err := service.UnmarshallUAST(&data)
+		if err != nil {
+			return nil, serializer.NewHTTPError(http.StatusBadRequest, err.Error())
+		}
+
+		resp := &uast.Node{InternalType: "Search results"}
+
+		if req.Filter != "" {
```
carlosms: Maybe an empty filter could be a bad request error, to simplify this part of the code.

server/handler/query.go
```diff
@@ -116,6 +116,7 @@ func Query(db service.SQLDB) RequestProcessFunc {
 					nodes, err := service.UnmarshallUAST(val)
 					if err == nil {
 						colData[columnNames[i]] = nodes
+						colData["__uast-protobufs"] = val
```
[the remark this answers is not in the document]
smacker: Great catch! thank you!

server/handler/uast.go
smacker: I'm open to this change. The only problem is then we will move complexity to frontend. Please let me know it's better to have it in JS than in Go.

carlosms: If you are talking about disabling the search button when the input is empty, I think it adds a nice feel to the interface and it's not that much complexity. I would make this change, but I'm also OK merging as it is. Up to you :+1:

smacker: I'm talking about how to return to full UAST.
Right now you can filter than just remove the filter query and get origin UAST back.
If we disable search button we would need one more button for it :) I personally use this feature.

smacker: I think for now I would keep current behavior to be consistent with dashboard. But feel free to create an issue if you see how to improve it! Thank you.

smacker:
rebased :cry: :cry: :cry:

smacker:
@bzz @carlosms please another pass

carlosms:
A few comments:

- UAST viewer has a loading indicator, but the CODE right panel does not. I would be good if this was consistent.
- If `/filter` fails the whole modal content gets replaced. this means that you cannot fix an xpath query with invalid syntax, for example. I think the UAST modal should use the same error overlay as the CODE modal.
a
- It would be better if an invalid xpath query returned an error different to 'internal server error'.

smacker:
issue to client-go: https://github.com/bblfsh/client-go/issues/65

PR #136, opened 2018-06-22
Issue 131

### LINKED ISSUE #131: Move UAST switcher in Code modal to the header
It might be tricky.
On the first glance without redux we would need HOC for the modal component to implement that.

carlosms:
Fixes #131.

It might be a good idea to do a similar refactor to `showUAST`, but it is unrelated and can be done when the code viewer is finished.

a

### DIFF
--- a/frontend/src/components/CodeViewer.js
+++ b/frontend/src/components/CodeViewer.js
@@ -6,14 +6,7 @@ import UASTViewer, { Editor, withUASTEditor } from 'uast-viewer';
 import api from '../api';
 import './CodeViewer.less';
 
-function EditorPane({
-  languages,
-  language,
-  showUast,
-  handleLangChange,
-  handleShowUastChange,
-  editorProps
-}) {
+function EditorPane({ languages, language, handleLangChange, editorProps }) {
   return (
     <div className="editor-pane">
       Language:{' '}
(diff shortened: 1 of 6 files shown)


frontend/src/components/CodeViewer.js
```diff
@@ -157,48 +137,73 @@ class CodeViewer extends Component {
...
+            CODE
+            <Switch
+              checked={showUast}
+              onChange={this.handleShowUastChange}
+              disabled={!language}
+              checkedIcon={<span className="switch-text checked">UAST</span>}
+              uncheckedIcon={
+                <span className="switch-text unchecked">UAST</span>
+              }
+              onColor="#ffba34"
```
smacker: do you think we can move it to css?

some class with `!important`.

frontend/src/components/CodeViewer.less
```diff
@@ -21,3 +23,22 @@
...
+.code-toggler {
+    margin-left: 40px;
+
+    .switch-text {
+        font-size: 18px;
+        font-weight: @btn-font-weight;
+    }
+
+    .switch-text.checked {
+        color:    @btn-gbpl-tertiary-color;
```
smacker: extra spaces

carlosms:
Rebased to fix conflicts.

I agree with the hardcoded colors in JS, but it's the only supported way by `react-switch`. I tried to move it to css in 42414e4, please take another look.

smacker:
thanks! yes. I prefer this `!important` override.

Issue #48, opened 2018-05-23
Show UAST from code


carlosms:
The details need to worked out, but the code viewer (#47) should be able to provide the UAST from the code, using the `/parse` endpoint (#46).

bzz:
The designs were updated in https://github.com/src-d/product/issues/264#issuecomment-398154008

carlosms:
This is a summary of the current understanding from IRL conversations, still open to more discussion and feedback from product team:

When you click on the table row 'CODE' button the code viewer modal opens.
At this moment the frontend calls `/detect-lang`.
- As a first iteration we will use the blob contents only.
- As an improvement, If the row contains the file name with the default column name, we may use it too. e.g. `file_name`, `tree_entry_name`.

Besides the automatic detection the modal will offer a drop-down with a list of languages, similar to the one in http://dashboard.bblf.sh/. If the language detection fails we should inform the user of the possibility to select a language manually.
If a language was detected this drop-down will be also enabled, indicating which language was automatically detected.

If a language is detected or selected, it will enable:
- Syntax highlighting
- The 'UAST' button in the top menu of the modal window.

Clicking on 'UAST' will open the right panel, possibly showing some kind of 'loading' message or animation.
The frontend will call `/parse` with the currently selected language, and fill the right panel with the UAST or error message.

carlosms:
Current status after the merge of #119.
Copied from https://github.com/src-d/gitbase-playground/pull/119#issuecomment-399035869:

Next steps:
1) show uast button is located in left pane not in the title of modal
2) there is no top bar on UAST (with locations, server, query) yet (#127).
3) bblfsh integration improvements (#123)
 - it doesn't return underlying error from bblfsh
 - list of languages is hard-coded
4) languages improvements
 - there are no workarounds for java/bash/??? yet (languageMode != language for them)
 - currently, we highlight only 6 languages which bblfsh knows. We can add more. Or not. 
5) frontend performance improvements

carlosms:
Most of the work is already in master.

@smacker do you have any comments about "frontend performance improvements"? Is it performing worse here than in the bblfsh dashboard? Are there any optimizations we can try?

smacker:
actually, it performs better for me now 😅  😅  😅

carlosms:
great! Then let's close this one and open new issues if we find any problems/improvements.

PR #140, opened 2018-06-27
Style for the modal contents

carlosms:
Part of #48.

There is a small difference with the design spec: the blue background of the code line gutter does not carry all the way to the top.
The reason is that the width is not fixed, it depends on the number of lines, it has different widths for 1 or 3 digit line numbers. There must be a convoluted way to make it work, but I think it's not worth it, at least for now.

a

b

c

### DIFF
--- a/frontend/src/App.less
+++ b/frontend/src/App.less
@@ -107,12 +107,6 @@ body {
 .modal-body {
     height: ~'calc(100vh - 150px)';
 
-    .uast-editor,
-    .react-codemirror2,
-    .CodeMirror {
-        height: 100%;
-    }
-
     .uast-viewer {
         height: 100%;
         overflow-y: auto;
--- a/frontend/src/App.less
+++ b/frontend/src/App.less
@@ -59,7 +59,7 @@ body {
         overflow: auto;
     }
 
-    .Resizer {
+    .Resizer.horizontal {
         border-top: solid 1px @primary-tint-2;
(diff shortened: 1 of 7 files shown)


frontend/src/App.less
```diff
@@ -85,12 +107,6 @@ body {
 .modal-body {
     height: ~'calc(100vh - 150px)';
 
-    .uast-editor,
```
smacker: are you sure we don't need it?

I remember one of the reasons to add it was "This site appears to use a scroll-linked positioning effect. This may not work well with asynchronous panning; see https://developer.mozilla.org/docs/Mozilla/Performance/ScrollLinkedEffects for further details and to join the discussion on related tools and features!".

frontend/src/variables.less
```diff
@@ -234,3 +254,10 @@
         }
     }
 }
+
+.btn-modal-close {
+    .btn-title-common;
+    .icon-btn-variant(@primary);
+
+    float: right;
+}
```
smacker: new line

frontend/src/App.less
smacker: nwm. I see you moved it to another file.

carlosms:
Thanks for the comments. I've checked and firefox does not complain in the console about the scroll, so it looks things work as they should.

Rebased and squashed the fixe for the new line. Let's merge :)

PR #153, opened 2018-06-28
Update client go

smacker:


carlosms:
Wait, if our go dependencies are already in `vendor`, why do we do a `dep ensure` in `back-dependencies`?

smacker:
yes Carlos, thanks! I'm merging then and create the [issue](https://github.com/src-d/gitbase-playground/issues/160). We had `dep ensure` before and it isn't related to client-go.

Issue #103, opened 2018-06-15
Define the help contents


carlosms:
The current design spec includes a 'help' button, but we still need to define:

- Behavior. Will this be a modal window? a tooltip? an external link?
- Contents.

Related to #44.

bzz:
Tagging @src-d/product here for feedback

bzz:
Ok, so the consensus is - it will be a link, pointing to https://docs.sourced.tech/gitbase 🎉

PR #156, opened 2018-06-28
add help link

smacker:
Fix: #103

It might look a bit strange when the first line of query is too long but I don't see how to make it better (from UI point of view).

screen shot 2018-06-28 at 16 51 02

### DIFF
--- a/frontend/src/components/QueryBox.js
+++ b/frontend/src/components/QueryBox.js
@@ -14,6 +14,7 @@ import 'codemirror/addon/hint/sql-hint';
 import './QueryBox.less';
 import SuccessIcon from '../icons/success-query.svg';
 import ErrorIcon from '../icons/error-query.svg';
+import HelpIcon from '../icons/help.svg';
 
 function ResultInfo({ result }) {
   if (!result) {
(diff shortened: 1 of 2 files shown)


carlosms:
I think that having that text hidden behind the button makes it look like something is a bit broken.

What do you think about shrinking the codemirror area, like this image? It will look the same with small queries that don't overflow.

image

smacker:
TBH I think it looks super ugly. Worse than hidden text.

@ricardobaeta could we have your input here? I don't think it's possible to make the text go around that button due to codemirror. So most probably we have only this 2 options as on my screenshot and as on Carlos screenshot.

ricardobaeta:
@carlosms @smacker 

Relying on the query input field scroll that we have, I propose just to give background: #fff to the container of the "HELP" element. With some nice padding.

smacker:
didn't get the idea. The original PR has `#fff` background. @carlosms maybe we should merge it then and let Ricardo fix it in browser?

ricardobaeta:
@smacker I like @carlosms suggestion. If we could have the padding a fill with gradient from #fff opaque and #fff transparent it would be a nice detail I think.

smacker:
sorry I still don't understand like at all :( gradient into the textarea with scroll? or how?

smacker:
got it.

Current look:
screen shot 2018-07-02 at 13 03 17

ricardobaeta:
Hi @smacker !

Nicely done ;)

carlosms:
Leaving aside my conviction that this looks like a bug instead of an intended design, I see a problem when I hover the icon. I guess the background should still be white to avoid the weird mix of texts.

a

smacker:
fixed hover

ricardobaeta:
Hi @carlosms!

You spotted a very nice detail, thank you ✨ 

I presume @smacker maintained the background on :hover, and I like it. For me it's fine.

Just food for thought:

image

PR #159, opened 2018-06-28
Shrink tab titles to fit in one line

### LINKED ISSUE #145: Improve tab titles layout for large number of tabs

carlosms:
This is my attempt to fix #145.

a

One possible improvement to make it easier to use with many tabs is to add tab change with the arrow keys or the mouse wheel. Actually the active tab is supposed to change with the arrow keys by default, but maybe the elements do not get focused or something...

Other improvement could be the addition of @ricardobaeta's drop down with the list of tabs next to history.

### DIFF
--- a/frontend/src/components/TabbedResults.js
+++ b/frontend/src/components/TabbedResults.js
@@ -1,5 +1,13 @@
 import React, { Component, Fragment } from 'react';
-import { Row, Col, Alert, Tab, Button } from 'react-bootstrap';
+import {
+  Row,
+  Col,
+  Alert,
+  Tab,
+  Button,
+  OverlayTrigger,
+  Popover
+} from 'react-bootstrap';
 import PropTypes from 'prop-types';
 import DivTabs from './DivTabs';
 import ResultsTable from './ResultsTable';
@@ -57,7 +65,7 @@ class TabTitle extends Component {
 
     if (inEdit) {
       return (
-        <div ref={this.ref}>
+        <div ref={this.ref} className="tab-title-wrapper">
           <input
             autoFocus
             type="text"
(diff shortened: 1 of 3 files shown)


ricardobaeta:
Hi @carlosms :) This is looking cool!

> One possible improvement to make it easier to use with many tabs is to add tab change with the arrow keys or the mouse wheel

I agree :)

> Other improvement could be the addition of @ricardobaeta's drop down with the list of tabs next to history.

Indeed! But I know that we're living a time to really make smart decisions, therefore I believe this solution won't hurt the experience that much. I propose only a little big details on tab :hover.

image

> This screen is updated in Zeplin

carlosms:
Branch rebased, there is a new commit to show the tooltip on hover. @smacker please take a look.

a

frontend/src/components/TabbedResults.js
```diff
@@ -74,21 +82,31 @@ class TabTitle extends Component {
...
-            this.props.handleRemoveResult(tabKey);
-          }}
-        />
-      </div>
+      <OverlayTrigger
+        placement="top"
+        delay={1000}
+        overlay={
+          <Popover id={`tooltip-${tabKey}`}>
+            {<span className="tab-tooltip">{title}</span>}
```
smacker: why do you need `{` `}` here?

frontend/src/components/TabbedResults.less
```diff
@@ -86,9 +86,16 @@
         flex: 0 0 auto;
         display: flex;
 
-        li {
+        li:not(.history-tab-title) {
             flex: 0 1 auto;
             overflow: hidden;
+
+            > div {
```
smacker: could you please use class name here?  It's hard to follow for which element you set this styles.

carlosms: This div is put by the bootstrap internals without any class name, I'd rather not customize it any more.

smacker: ah. I see now why you do it.

carlosms:
Pushed a new commit with the requested changes

a

smacker:
Awesome job!

ricardobaeta:
@carlosms Nicely done ;)

Just a minor detail, about the vertical alignment of the query text. I would love for it to vertical align center with the height of the tooltip.

image

carlosms:
I created an issue so we don't forget about this https://github.com/src-d/gitbase-playground/issues/166

Issue #161, opened 2018-06-29
Remove the custom bblfsh server input


carlosms:
As discussed [here](https://github.com/src-d/product/issues/264#issuecomment-401102690).

PR #163, opened 2018-06-29
Remove server input

smacker:
Based on #135

Fix: #161

Custom server removed only from frontend as requested by the issue. If you think it should be removed from backend too, please let me know.


smacker:
rebased

PR #164, opened 2018-06-29
change default gitbase user to root

smacker:
gitbase v0.14.0-rc1 changed user from `gitbase` to `root`.

Travis is currently failing on master: https://travis-ci.org/src-d/gitbase-playground/jobs/398254777


smacker:
I updated helm config here too. Otherwise after merge staging will be broken.

smacker:
the last comment isn't true. Chart sets user for both gitbase & playground. So it should work with any version of gb. I removed the commit with update.

Issue #165, opened 2018-07-02
Add configurable small HTML snippet to the frontend


bzz:
Gitbase needs to provide a user with ability to make small, per-deployment customizations of the frontend.

This feature targets adding non-visual HTML tags, with a main use case of adding common web Analytics snippet to every page, in case a user may want to do so, on one of his own playground deployments.

As a standalone application, gitbase-playground codebase will not include any user tracking, or anything like that.

Simplest way could be adding something like `FOOTER_HTML` ENV VAR configuration option, so that

```sh
$ tee -a ga-snippet.txt << END
<!-- Google Analytics -->
<script>
window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
ga('create', 'UA-XXXXX-Y', 'auto');
ga('send', 'pageview');
</script>
<script async src='https://www.google-analytics.com/analytics.js'></script>
<!-- End Google Analytics -->
END

$ cat ga-snippet.txt | base64
$ export FOOTER_HTML="PCEtLSBHb29nbGUgQW5hbHl0aWNzIC0tPgo8c2NyaXB0Pgp3aW5kb3cuZ2E9d2luZG93LmdhfHxm
dW5jdGlvbigpeyhnYS5xPWdhLnF8fFtdKS5wdXNoKGFyZ3VtZW50cyl9O2dhLmw9K25ldyBEYXRl
OwpnYSgnY3JlYXRlJywgJ1VBLVhYWFhYLVknLCAnYXV0bycpOwpnYSgnc2VuZCcsICdwYWdldmll
dycpOwo8L3NjcmlwdD4KPHNjcmlwdCBhc3luYyBzcmM9J2h0dHBzOi8vd3d3Lmdvb2dsZS1hbmFs
eXRpY3MuY29tL2FuYWx5dGljcy5qcyc+PC9zY3JpcHQ+CjwhLS0gRW5kIEdvb2dsZSBBbmFseXRp
Y3MgLS0+Cg=="

$ gitbase-playground 
```

adds a snippet to every page.

smacker:
maybe it's better to use file instead of base64 env var?

There is no problem to use files with k8s. Example:
```
kind: ConfigMap
metadata:
  name: my-configs
data:
  some.file.txt: |-
    <content>

kind: Deployment
spec:
  template:
    containers:
      ...
      volumeMounts:
        - name: config-volume
        - mountPath: /etc/my-app/
  volumes:
[...]

PR #171, opened 2018-07-02
Add configurable small HTML snippet to the frontend

smacker:
Fix: #165 

I ignore error if a wrong base64 was provided. Please let me know if you think we need to check it on the start of the server and doesn't allow it to start/write error somewhere.

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -37,5 +37,7 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
+    <!-- create react app removes comments during build that's why we need div -->
+    <div class="invisible-footer"></div>
   </body>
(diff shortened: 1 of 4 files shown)


frontend/public/index.html
```diff
@@ -37,5 +37,7 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
+    <!-- create react app removes comments during build that's why we need div -->
+    <div class="invisable-footer"></div>
```
carlosms: Typo, should be invisible.

smacker:
typo fixed. doc updated. though it's very strange that we put them in CONTRIBUTING...

carlosms:
Yes, I agree, I'd like to improve the documentation because all the information is scattered in too many places.

Issue #173, opened 2018-07-03
Add Integration Test to CI to check presence of FOOTER_HTML


bzz:
Add curl to CI to check HTML snippet was added by https://github.com/src-d/gitbase-playground/issues/165

As discussed elsewhere: `curl` + `grep` should be enough for this case.

carlosms:
Maybe someone can help me out.

I tried to do a test in my fork:
https://github.com/carlosms/gitbase-playground/commit/ad2f4908e9d56ba29c9dcd0ef73751315f0c2e96

But I don't see the reason why it works locally but fails in Travis:
https://travis-ci.org/carlosms/gitbase-playground/jobs/399583173#L965

Any ideas?

smacker:
As suggested in slack removing last slash helped:
https://github.com/src-d/gitbase-playground/pull/177
https://travis-ci.org/src-d/gitbase-playground/jobs/399935192

carlosms:
After doing some more testing, the test always fails for my fork. But a custom trigger with the same travis conf in this project master works with slashes, localhost, 127.0.0.1...
https://travis-ci.org/src-d/gitbase-playground/builds/399945172

PR #177, opened 2018-07-04
Add travis test for FOOTER_HTML

smacker:
Fix: #173

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,15 @@ jobs:
       - sleep 15
       - GITBASEPG_DB_CONNECTION='root@tcp(localhost:3367)/none' GITBASEPG_INTEGRATION_TESTS=true make test
     - script:
-        - make packages
+      - make packages
+    - script:
+      - PKG_OS=linux make build
+      - make docker-build
+      - export FOOTER_HTML='<div class="custom-base64-html"></div>'
+      - export FOOTER_HTML_BASE64=`echo -n $FOOTER_HTML | base64`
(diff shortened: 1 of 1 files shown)


smacker:
Drone has failed on cloning repo.

bzz:
Nice!

Suggestion - what do you think, if instead of using hard-coded base64 value for `FOOTER_HTML` we could use 2 explicit shell command to generate one from a small plain text HTML snippet, as i.e suggested in https://github.com/src-d/gitbase-playground/issues/165 ? 

I think that could be easier to understand and modify it later, compared to already encoded value.

.travis.yml
```diff
@@ -52,7 +52,14 @@ jobs:
...
     - script:
-        - make packages
+      - make packages
+    - script:
+      - PKG_OS=linux make build
+      - make docker-build
+      - docker run --rm -d --name gitbasepg --publish 8080:8080 --env GITBASEPG_FOOTER_HTML=PGRpdiBjbGFzcz0iY3VzdG9tLWJhc2U2NC1odG1sIj48bWFycXVlZT7Cr1xfKOODhClfL8KvPC9tYXJxdWVlPjwvZGl2Pgo= srcd/gitbase-playground:$TRAVIS_BRANCH
+      - sleep 5
+      - curl http://127.0.0.1:8080
+      - curl http://localhost:8080 | grep '<div class="custom-base64-html">'
```
bzz: Would 1 curl be not enough here?

PR #188, opened 2018-07-10
Show limit

### LINKED ISSUE #117: Inform the user of the hardcoded LIMIT
We add a hardcoded `limit` to every query, but it's not obvious to the user.
When you realize there is missing data it looks like something is broken.

My proposal is to return the edited query and display that inside the results tab. So for the query
`SELECT * FROM files`

You would get on top of the results table
`SELECT * FROM files LIMIT 100`

cc @src-d/product team.

carlosms:
Fix #117.

This PRs adds a small message with the forced limit near the table pagination controls:
a

ping @ricardobaeta, in case you want to take a look.


ricardobaeta:
Nicely done ;)

PR #190, opened 2018-07-10
Use build tags to manage go-bindata

### LINKED ISSUE #24: Backend: Rethink how to work with assets, bindata...
from https://github.com/src-d/gitbase-playground/pull/15#discussion_r186789669 @smacker suggested that it would be great to:
> _[...] re-think how we work with assets. Maybe we can find better solution._

carlosms:
My proposal to fix #24.

With this PR there are 2 static files handlers with conditional compilation:
- `static_files_dev.go`: always returns an error, "Frontend assets are only available when using 'make build' or 'make serve'"
- `static_files.go`: uses `server/assets/asset.go`, that needs to be created with `go-bindata`.

The default one is `_dev`. The `go-bindata` one is used when `-tags bindata` is used in `go run/build`.

From the developers and users point of view nothing changes. You still need to build the frontend with `make serve`, just as before. `make build/package` remain the same.


carlosms:
Did a small fix, ready for review now.

PR #191, opened 2018-07-10
Fix value for float columns

### LINKED ISSUE #189: SUM result interpreted as UAST.
This query result is interpreted as a UAST:
`select SUM(blob_size/1024) from blobs;`

carlosms:
Fix #189.

The first commit is just a refactor and tests, the fix is in the second one.

### DIFF
--- a/server/handler/query.go
+++ b/server/handler/query.go
@@ -146,6 +146,11 @@ func columnsData(
 			if sqlVal.Valid {
 				colData[columnNames[i]] = sqlVal.Int64
 			}
+		case *sql.NullFloat64:
+			sqlVal, _ := val.(*sql.NullFloat64)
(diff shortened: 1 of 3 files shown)


carlosms:
Mapping it to a boolean makes the output of things like `is_remote(ref_name)` a nice `true/false`. Otherwise it will be `1/0`. Which would be technically correct, of course, but I assume less desirable.

@src-d/data-retrieval is there any risk of encountering a column type `BIT` with something that is not a boolean 1/0?

erizocosmico:
Nope, we use `BIT` for boolean inside go-mysql-server.

carlosms:
Thanks @erizocosmico! Let's merge.

PR #195, opened 2018-07-11
Keep tabs open on error

### LINKED ISSUE #185: Results tab disappears if query fails
Right now, if a new query fails after some time of execution (~30sec) - error message is shown, but the tab disappear 😕 

That is inconsistent with the rest of the tab behaviour. To be consistent, we should keep the tab open, it should have empty results and an some kind of "error" state of a "broken robot".

@ricardobaeta what do you think?

Example of the query that fails slow in details below.
<details>

```sql
SELECT files.repository_id, files.file_path,
 ARRAY_LENGTH(UAST(
   files.blob_content,
   LANGUAGE(files.file_path, files.blob_content),
   '//*[@roleFunction and @roleDeclaration]'
 )) as functions
FROM files
NATURAL JOIN refs
WHERE 
LANGUAGE(files.file_path,files.blob_content) = 'Go'
AND refs.ref_name = 'HEAD'
ORDER BY functions
```
</details>

carlosms:
Fix #185.

There is a difference in how chrome and firefox render the gif. Chrome plays the animation only one the first time :/

Chrome:
a

Firefox:
b

Maybe we could modify the gif to be a loop, with the last frame paused for 5 minutes. What do you think @ricardobaeta, @bzz, @smacker ?


smacker:
@carlosms there is no much problem with restarting the animation. Something like this would work if you want to keep the same element: https://jsfiddle.net/ko4stxwf/4/

But with react you most probably can create new instance of img and it should work fine (most probably, I didn't test)

ricardobaeta:
@carlosms @bzz @smacker 

There's no worries creating the 5min paused loop. But i think the size of the asset will go up. Would Maxim's suggestion be better?

carlosms:
We saw 2 options, doing the img src change with a timeout, or adding `?xx` to each image src.
Both are not ideal, so I went with the second one 418d206

Issue #184, opened 2018-07-05
Version endpoint should include versions of dependencies


bzz:
Right now we only display version of the gibase-playground build.

This issue is about adding 2 extra fields:
 - bblfsh version
 - gitbase version

in the output of the `curl <host>/version`

carlosms:
gitbase has a version command, but that's not really accessible to us, we only have the DB connection endpoint.

Should we open a request for them to implement [`VERSION`](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_version)?

smacker:
VERSION function is implemented in gitbase master. Let's wait for new release?

carlosms:
Or we could implement it now against master, reporting the version as unknown if `VERSION` is not available.

Are we going to include this information in the interface somewhere, similar to bblfsh dashboard?

bzz:
@carlosms for now, let's include it in only in `/version` endpoint results

>  reporting the version as unknown if VERSION is not available.

sounds good!

PR #196, opened 2018-07-16
Version endpoint should include versions of dependencies

smacker:
Fix: #184

### DIFF
--- a/server/handler/version.go
+++ b/server/handler/version.go
@@ -13,21 +13,24 @@ import (
 // with a current version of server and dependencies
 func Version(version, bbblfshServerURL string, db service.SQLDB) RequestProcessFunc {
 	return func(r *http.Request) (*serializer.Response, error) {
-		cli, err := bblfsh.NewClient(bbblfshServerURL)
-		if err != nil {
-			return nil, err
-		}
-
-		resp, err := cli.NewVersionRequest().Do()
-		if err != nil {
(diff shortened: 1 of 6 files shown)


server/handler/version.go
```diff
@@ -3,13 +3,31 @@ package handler
...
 	return func(r *http.Request) (*serializer.Response, error) {
-		return serializer.NewVersionResponse(version), nil
+		cli, err := bblfsh.NewClient(bbblfshServerURL)
+		if err != nil {
+			return nil, err
+		}
+
+		resp, err := cli.NewVersionRequest().Do()
+		if err != nil {
+			return nil, err
```
carlosms: So this will return an error if bblfsh is not reachable? What do you think about doing the same as gitbase, returning "unknown"? This way in the interface you could quickly see if one of the services is down.

smacker: I thought about it but wasn't sure it's correct behavior. But if you think the same, I'm changing.

carlosms:
LGTM

smacker:
Test code coverage is reduced a little. But I'm not sure it worths to write complicated tests with mocked bblfsh & gitbase for this endpoint.

carlosms:
I agree mocking that would be too much trouble. If we really want to we can make an integration test with the real services, a simple one that checks that the version is not "undefined".

Issue #183, opened 2018-07-05
Version endpoint should identify the exact source code revision


bzz:
We have an endpoint to see the version of the gitbase-playground i.e at 

```
$ curl https://gitbase-playground-staging.srcd.run/version
```

but it is ambiguous. It should identify the exact source code revision used for the build./

Expected outcome:
 - an identifier that points to the uniq revising of the code, used for this build i.e 
   `v0.1.0`  or  it's sha1 `9a885b0`

Current outcome:

 - ```
    {"status":200,"data":{"version":"master"}}
    ````

With current behaviour it is impossible to identify the revision i.e to get the same version locally and reproduce a bug.

carlosms:
I created a PR in the CI repo for this: https://github.com/src-d/ci/pull/82

There  is a minor problem. The version name for this project will have the `-dirty` suffix when it's not a tag build. This is because of how we do go-bindata, it makes changes to a file in git (see #24).

bzz:
82 is merged now, what would be the next steps?

carlosms:
We just need to force a new deployment doing a new commit to master.
Conveniently, I just found a bug in the makefile, #201, so we can use that one :D

bzz:
Shall we merge #201 and deploy?

carlosms:
#201 merged.
We still need to:
- update gitbase to 0.15.0 (#204) to report its version
- fix #203, until then there are changes after `dep ensure` and the version reported has the suffix `-dirty`.

bzz:
Great, all looks ready to be merged and :shipit: !

bzz:
https://gitbase-playground-staging.srcd.run/version
```
{"status":200,"data":{"version":"dev-558c67b","bblfsh":"v2.5.0","gitbase":"8.0.11-undefined"}}
```
🎉

PR #201, opened 2018-08-10
Fix go build args evaluation in makefile

carlosms:
Part of #183.
`LD_FLAGS` was empty, with this PR it is evaluated when build is called and not when parsing the file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,9 +27,6 @@ COMPOSE := docker-compose
 # To be used as -tags
 GO_BINDATA_TAG := bindata
 
-# Override Makefile.main defaults for arguments to be used in `go` commands.
-GO_BUILD_ARGS := -ldflags "$(LD_FLAGS)" -tags "$(GO_BINDATA_TAG)"
-
 # Environment and arguments to use in `go run` calls.
 GO_RUN_ENV := GITBASEPG_ENV=dev
 GO_RUN_ARGS += -tags "$(GO_BINDATA_TAG)"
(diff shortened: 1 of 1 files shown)


bzz:
TravisCI seems not healthy today
```
docker run --rm -d --publish 8080:8080 --env GITBASEPG_FOOTER_HTML=$FOOTER_HTML_BASE64 srcd/gitbase-playground:$TRAVIS_BRANCH

Unable to find image 'srcd/gitbase-playground:master' locally
docker: Error response from daemon: manifest for srcd/gitbase-playground:master not found.
See 'docker run --help'
```

or 


```
0.41s$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends $TRAVIS_APT_OPTS install gcc-6 g++-6
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package gcc-6
E: Unable to locate package g++-6
E: Couldn't find any package by regex 'g++-6'
```

smacker:
Good.
But @carlosms  could you please explain why it must be defined inside build not at the top of the file? I don't really understand and can make the same mistake because of it later.

carlosms:
The problem was that `GO_BUILD_ARGS` used the still unset `LD_FLAGS`, which didn't fail, but was put as a empty string.

`LD_FLAGS` is set in the `Makefile.main` from the `.ci` folder, which was not imported at the very top, it is done a few lines later:
```
MAKEFILE := $(CI_PATH)/Makefile.main
$(MAKEFILE):
	@git clone --quiet --depth 1 -b $(CI_BRANCH) $(CI_REPOSITORY) $(CI_PATH);
-include $(MAKEFILE)
```

Instead of moving `GO_BUILD_ARGS` below this include, I thought it made sense to define it later for the only rule that uses it.

smacker:
ahhh. Yeah. I noticed it was defined too early. So there is no requirement to always put it into the target. Got it. Thanks!

carlosms:
The CI build fails because of #202 and #203.

I'll merge this one as those 2 are unrelated to the changes here, and the rest of the tests pass.

Issue #160, opened 2018-06-29
dep ensure in back-dependencies


smacker:
Do we need it? Or we can we remove it?
We vendor all dependencies, so it isn't necessary to run `dep`. Do I miss some use case?

bzz:
AFAICS  [back-dependencies](https://github.com/src-d/gitbase-playground/blob/master/Makefile#L95) runs only [dep ensure](https://golang.github.io/dep/docs/daily-dep.html#basics)  and is used only in `dependencies` and `validate-commit`.

At least for dev workflow, I see a value in `make dependencies` in the case of a new lib added to the backend as it takes care of a new dependency, and make sure it's included in `.toml`.

smacker:
@bzz could you please provide a particular use case?

For now, I see that our deps are vendored so `dep ensure` must do nothing. If it changes anything it's wrong. And to check that we vendored our deps correctly we have `validate-commit`.

carlosms:
Closing, it was removed in #205.

Issue #202, opened 2018-08-13
Wrong local docker image tag in Travis tests


carlosms:
After we merged https://github.com/src-d/ci/pull/82, the local docker image is tagged with `$(DEV_PREFIX)-$(COMMIT)$(GIT_DIRTY)` or `TRAVIS_TAG`.

In [travis.yml](https://github.com/src-d/gitbase-playground/blob/master/.travis.yml#L61) we still use the docker image `srcd/gitbase-playground:$TRAVIS_BRANCH`, and builds fail.

bzz:
Thank you for prompt fixes!

PR #205, opened 2018-08-13
 Makefile uses go dep releases instead of master

### LINKED ISSUE #203: Update go dep lock file, use latest release instead of master
Go dep has been updated and `make validate-commit` fails because of differences in the `Gpkg.lock` file.

While we are at it, we could install the latest release instead of master in travis.
If we copy the changes from https://github.com/src-d/ci/pull/84 to our local Makefile we can remove them later when that PR is merged into `.ci/Makefile.main`.

carlosms:
Fix #203.

I also went ahead and removed `back-dependencies`, because I agree with @smacker in #160.
The dependencies are already vendored and we don't need to update them every single time we build. Now we can do a `make godep` whenever we know we need to update the dependencies.

But I will revert this last point with no problems if @bzz feels strongly about it.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,7 +2,7 @@
 
 
 [[projects]]
-  digest = "1:e0499b93857a1b91efbdc0c21e26f1130685de98035008f3b01eeeb0713798cb"
+  digest = "1:167b6f65a6656de568092189ae791253939f076df60231fdd64588ac703892a1"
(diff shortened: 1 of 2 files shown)


carlosms:
CI is failing because of #202, the tests related to these changes are OK: https://travis-ci.org/src-d/gitbase-playground/jobs/415410597#L1187

Issue #210, opened 2018-08-14
Gracefully remove the container when Ctrl-C


eiso:
When I run gitbase via docker, and Control-C the server, it gracefully exits and removes the container from docker engine. When I do the same on the playground, I still have to manually remove it with `docker rm -f gitbase-playground`, I don't think there is any need to persist the container, I would remove it like we do with gitbase or bblfsh.

bzz:
I belive using `docker-compose down` after `ctrl-c` is the solution for this issue.

If that works, it shall be documented in https://github.com/src-d/gitbase-playground#with-docker-compose

eiso:
I meant without docker-compose. I don't know how gitbase does it, but it cleans up the exited container.

bzz:
@eiso if you want a single container to be deleted, you need to add `--rm`  [option to the run command](https://docs.docker.com/engine/reference/run/#clean-up---rm).

So it shall be 
```
docker run -it --rm \
    --name gitbase-playground \
    --network host \
    --env GITBASEPG_DB_CONNECTION="root@tcp(127.0.0.1:3306)/none" \
    --env GITBASEPG_SELECT_LIMIT=0 \
    srcd/gitbase-playground:latest
```

eiso:
Of course! [insert facepalm gif] [insert embarrassed gif]

bzz:
N.p!

bzz:
Closing as issue seems to be resolved and #213 is merged

PR #213, opened 2018-08-14
Doc: add cleanup instructions for docker-compose scenario

bzz:
Solves #210

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,12 @@ $ GITBASEPG_REPOS_FOLDER=$HOME/repos docker-compose up --force-recreate
 
 The server should be now available at [http://localhost:8080](http://localhost:8080).
 
+In case there are any containers left, you can use
+```bash
+docker-compose down
+```
+for cleanup.
+
 ## Without Docker Compose
 
 The playground will run the queries against a [gitbase](https://docs.sourced.tech/gitbase) server, and will request UASTs to a [bblfsh](https://doc.bblf.sh/) server. Make sure both are properly configured.


carlosms:
If I'm not mistaken this command would only apply if we started the containers with `up -d`, which is not the case.

Following the instructions from the readme the `up` command stays in the foreground and Ctrl+C is enough to stop it, which should be obvious to anyone familiar with the shell.

carlosms:
Oh, and I've just realized that we do indeed already mention it:
> To kill the running containers use `Ctrl+C`.

bzz:
@carlosms I would keep it as it's useful in case:

 - using https://github.com/src-d/gitbase-playground#with-docker-compose
 - user runs `docker-compose up`
 - but there already is bblfsd container on his machine on the same port
 - so the process exists, leaving some containers running, some not (lookout and bblfsh shutdown, analyzer and db still running)

So reminder of using `docker-compose down` to clean things up may be nice to have.

carlosms:
I see your point, but at the same time from my point of view that's troubleshooting and not normal usage, and it is written as a required step after every usage.

I just think it's not needed, but I'm not strongly against it. Please go ahead and merge if you consider it's useful.

README.md
```diff
@@ -33,6 +33,12 @@ $ GITBASEPG_REPOS_FOLDER=$HOME/repos docker-compose up --force-recreate
 
 The server should be now available at [http://localhost:8080](http://localhost:8080).
 
+In case if there are any containers left, you can use
```
carlosms: I believe this needs to be **In case there are** or **If there are**

bzz:
Thank you @carlosms ! History re-written

PR #217, opened 2018-08-16
Cancel sql requests from the UI

### LINKED ISSUE #199: When stopping a query, kill the query in gitbase server
When stopping the execution of a query by pressing (x) in the UI, we should be able to kill that query on the server side. Right now, it seems that query will continue to execute. This probably requires an issue in the server project as well.

carlosms:
Fix #199.

The abort-controller polyfill is used only for the jest tests. All major browsers apparently support AbortCotroller (https://caniuse.com/#search=fetch), and for example for older IE this PR does not break compatibility because it doesn't support fetch API either, which we already use.

I refactored the suspended tab reload to act more like a new query submit. This way the history also shows a new entry for the reload queries. In master reload actions are not reflected in the history.

Tested for chrome and firefox, please let me know if safari works as expected.

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -59,6 +59,7 @@ class App extends Component {
     this.handleTextChange = this.handleTextChange.bind(this);
     this.handleSubmit = this.handleSubmit.bind(this);
     this.handleRemoveResult = this.handleRemoveResult.bind(this);
+    this.handleAbortQuery = this.handleAbortQuery.bind(this);
     this.handleTableClick = this.handleTableClick.bind(this);
     this.handleExampleClick = this.handleExampleClick.bind(this);
     this.handleModalClose = this.handleModalClose.bind(this);
@@ -209,12 +210,12 @@ AND refs.ref_name = 'HEAD'`
     });
   }
 
-  handleSubmit() {
-    const { sql, history } = this.state;
-    const key = nanoid();
+  loadQuery(key, sql) {
+    const { history } = this.state;
+    const abortController = new window.AbortController();
 
     const loadingResults = new Map(this.state.results);
-    loadingResults.set(key, { sql, loading: true });
+    loadingResults.set(key, { sql, loading: true, abortController });
 
     this.setState(
       {
@@ -234,11 +235,11 @@ AND refs.ref_name = 'HEAD'`
     );
 
     api
-      .query(sql)
+      .query(sql, abortController.signal)
       .then(response => {
         this.setResult(key, { sql, response });
 
-        // If the schema of the list of languages were not loaded for some
+        // If the schema or the list of languages were not loaded for some
         // reason, we retry now
 
         if (!this.state.schema) {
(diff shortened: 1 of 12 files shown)


server/handler/query.go
```diff
@@ -50,6 +51,9 @@ func genericVals(colTypes []string) []interface{} {
 // the rows as JSON
 func Query(db service.SQLDB) RequestProcessFunc {
 	return func(r *http.Request) (*serializer.Response, error) {
+		ctx, cancel := context.WithCancel(r.Context())
```
smacker: why do you need to wrap it in WithCancel?

server/handler/query.go
```diff
@@ -62,41 +66,60 @@ func Query(db service.SQLDB) RequestProcessFunc {
...
-		}
-		defer rows.Close()
+		respCh := make(chan *serializer.Response)
+		errCh := make(chan error)
 
-		columnNames, columnTypes, err := columnsInfo(rows)
-		if err != nil {
-			return nil, err
-		}
+		go func() {
```
smacker: could you please explain why do we need to wrap it in goroutine?

server/handler/query.go
carlosms: It's true, we don't need the derived context, because we are not calling cancel anywhere.
I think I saw a couple of examples where they did this, but it was probably to accommodate a more generic scenario where you want to cancel early for some reason.

Should we remove it and use the request context instead?

server/handler/query.go
carlosms: My intention was to wrap all the logic to be able to cancel not only the db query, but also the potentially costly processing on the results. But, reading it again, I realize it's no use because the code would need to listen for the context cancellation in more places.

I'll remove the goroutine and simplify to 
```
		rows, err := db.QueryContext(ctx, query)
		if err != nil {
			return nil, err
		}
```

bzz:
Network failed on Drone
```
+ git fetch --no-tags origin +refs/pull/217/merge:
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
fatal: The remote end hung up unexpectedly
exit status 128
```

frontend/src/App.js
```diff
@@ -234,11 +235,11 @@ AND refs.ref_name = 'HEAD'`
     );
 
     api
-      .query(sql)
+      .query(sql, abortController.signal)
       .then(response => {
-        this.setResult(key, { sql, response });
+        this.setResult(key, { sql, response, abortController });
```
smacker: why do you need abort controller after request is finished already?

frontend/src/App.js
```diff
@@ -249,9 +250,26 @@ AND refs.ref_name = 'HEAD'`
...
         }
       })
-      .catch(msgArr =>
-        this.setResult(key, { sql, errorMsg: msgArr.join('; ') })
-      );
+      .catch(msgArr => {
+        this.setResult(key, {
+          sql,
+          errorMsg: msgArr.join('; '),
+          abortController
```
smacker: same here. API returned error. Looks like abort doesn't make sense here

frontend/src/App.js
smacker: I would really prefer to remove it. You have a check for the existence of abortController in cancel function. I spent quite a while to find for which case it might be needed but apparently it's not. I think removing redundant code makes it easier to read & understand.

carlosms: done in 359d4de

Issue #218, opened 2018-08-17
Rename src-d/gitbase-playground to src-d/gitbase-web


bzz:
- [x] Change repository name. (related: #223)
- [x] Travis CI sync
- [x] Update README/CONTRIBUTING. #220.
- [x] Rename binary [`gitbase-playground`](https://github.com/src-d/gitbase-playground/tree/master/cmd/gitbase-playground) -> `gitbase-web` #221 & #225.
- [ ] Check links from documentation of other projects.
- [x] Create srcd/gitbase-web on Docker Hub.
- [x] Push older images to srcd/gitbase-web.
- [x] Cur a release: push images to srcd/gitbase-web Docker Hub. [v0.1.3](https://github.com/src-d/gitbase-web/releases/tag/v0.1.3)
- [x] Update srcd/gitbase-playground description on Docker Hub to point to srcd/gitbase-web.
- [x] Coordinate update of the [src-d/charts/gitbase-playground](https://github.com/src-d/charts/tree/master/gitbase-playground). #222 & https://github.com/src-d/charts/pull/65
- [x] Setup drone for the new repository name https://github.com/src-d/issues-infrastructure/issues/214
- [x] Decide if we want to change the staging url, and open an issue for it.

Before doing any of the changes, let's double check that the task list is comprehensive.

bzz:
@smola can help handing administrative part of renaming.

@carlosms could you please help with other parts of it?

carlosms:
:+1: Off the top of my head I don't see anything missing.

carlosms:
I updated the description to include links to each PR.

Regarding "Check links from documentation of other projects". I couldn't find any links to this repository in any other projects. Please @bzz also take a look in case I missed something.

bzz:
> Regarding "Check links from documentation of other projects".

@carlosms I belive it's https://docs.sourced.tech/intro/#getting-started-with-gitbase-and-gitbase-playground but IDK which repository is that hosted in. 

@eiso could you please give us a hand here?

carlosms:
Linking a question by @rporres for visibility here:
https://github.com/src-d/gitbase-web/pull/222#issuecomment-413908730
[Regarding staging url]
> I don't know if you want to change the name or not, most likely yes, but I have received no issue around this. So before I approve this change I would need to know what you want to do with the gitbase playground domain

eiso:
@bzz I'll have a look through the documentation myself.

carlosms:
[...]

PR #221, opened 2018-08-17
Rename to Web: binary and docker images

carlosms:
Part of #218.

I think we should be able to merge this without depending for the repo name change.


docker-compose.build.yml
```diff
@@ -1,6 +1,6 @@
 version: '3.3'
 
 services:
-  playground:
-    image: gitbase-playground-dev
+  web:
```
bzz: Here and below, would `web-client` be a bit more descriptive?

carlosms: Let's call it gitbase-web, even more descriptive

carlosms: Done in 04407f3

PR #222, opened 2018-08-17
Rename to Web: helm chart

carlosms:
Part of #218.

Depends on https://github.com/src-d/charts/pull/65.

@rporres: will we make any changes to these two values?
```
ingress.globalStaticIpName=gitbase-playground-staging,
ingress.hostname=gitbase-playground-staging.srcd.run
```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -30,7 +30,7 @@ helm_deploy: &helm_deploy
   image: quay.io/ipedrazas/drone-helm:master-9b37211
   skip_tls_verify: true
   helm_repos: srcd-charts=https://src-d.github.io/charts/
-  chart: srcd-charts/gitbase-playground
+  chart: srcd-charts/gitbase-web
   release: gp
   tiller_ns: kube-system
   wait: true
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -63,7 +63,7 @@ pipeline:
     <<: *helm_deploy
     prefix: STG
     secrets: [ STG_API_SERVER, STG_KUBERNETES_TOKEN ]
-    values: gitbasePlayground.image.tag=commit-${DRONE_COMMIT_SHA:0:7},ingress.globalStaticIpName=gitbase-playground-staging,ingress.hostname=gitbase-playground-staging.srcd.run
+    values: gitbaseWeb.image.tag=commit-${DRONE_COMMIT_SHA:0:7},ingress.globalStaticIpName=gitbase-playground-staging,ingress.hostname=gitbase-playground-staging.srcd.run
```
smacker: we didn't update helm chart yet, right?
Maybe don't change it as long as helm chart expects `gitbasePlayground`?

carlosms: yes, this PR depends on https://github.com/src-d/charts/pull/65 being merged. Actually the only thing this PR does is use that chart rename, we don't need to merge anything until then.

rporres:
those changes imply much more than it looks as google deployments are slow... 

I don't know if you want to change the name or not, most likely yes, but I have received no issue around this. So before I approve this change I would need to know what you want to do with the gitbase playground domain

carlosms:
@rporres do you think we could merge https://github.com/src-d/charts/pull/65 and this one as they are, and let us consider if we want to change the domain name later on?

PR #229, opened 2018-08-29
quickstart.md: remove bogus instruction

mpl:
The "playground" service does not exist in docker-compose.yml, so the instruction to pull it was failing. In addition, it seems that 'docker-compose up' automatically pulls all the necessary images anyway, so there's no need to 'docker-compose pull' anything beforehand.


carlosms:
Hi @mpl, thanks for contributing!

But this is not an active branch, and the contents are a bit outdated (the project was renamed from Playground to Web).

This was created as a work in progress for unpublished gitbook docs, apologies if you found a link to it somewhere.
For now please use the master README as the source of documentation for this project.

PR #254, opened 2018-10-04
Update to bblfsh 2.9, use -drivers image

carlosms:
- Use bblfsh `-drivers` image in `docker-compose` and dev. docs.
- Bring back the `latest` bblfshd to travis to detect problems with new releases.
- Update staging to 2.9.0

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,16 +19,5 @@ services:
     volumes:
       - ${GITBASEPG_REPOS_FOLDER}:/opt/repos
   bblfsh:
-    image: "bblfsh/bblfshd:v2.7.2"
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -19,16 +19,8 @@ services:
...
-    volumes:
-      - type: volume
-        source: drivers
-        target: /var/lib/bblfshd
-    entrypoint: ["/bin/sh"]
-    command:
-    - "-c"
-    - "bblfshd & sleep 5 && bblfshctl driver install --recommended && tail -f /dev/null"
 
 volumes:
```
smola: No need for the volumes section anymore?

carlosms: You are right, thanks. Fixed in f5224d1

PR #256, opened 2018-10-04
Fix UAST unmarshal for empty strings

### LINKED ISSUE #255: Some row contents are incorrectly shown as UAST
In staging the 'Top languages by repository count' query creates a UAST button for a cell that has no UAST contents.

carlosms:
Fix #255.

For empty strings `UnmarshalUAST` does not return an error, it returns `nil, nil`.


carlosms:
Test added, squashed into one commit.

carlosms:
I forgot about the export handler :/
Second commit contains a fix and test for it also.

PR #261, opened 2018-10-05
[HOTFIX] Set DB conn reuse timeout

carlosms:
After an idle time of >60s the next query was always failing with `Query Failed - invalid connection`.
Apparently the problem is that the connection was close on gitbase's end and we still kept it open on our side.

More context in this thread:
https://github.com/go-sql-driver/mysql/issues/674#issuecomment-345661869

### DIFF
--- a/cmd/gitbase-web/main.go
+++ b/cmd/gitbase-web/main.go
@@ -51,6 +51,8 @@ func (c *ServeCommand) Execute(args []string) error {
 	}
 	defer db.Close()
 
+	db.SetMaxIdleConns(0)
+
 	static := handler.NewStatic("build/public", c.ServerURL, c.SelectLimit, c.FooterHTML)
 
 	// start the router


ajnavarro:
I think tis will not solve the issue. The connection will expire one minute after it was first created, I guess causing the same error.

ajnavarro:
I think you should use https://golang.org/pkg/database/sql/#DB.Ping before try to use any of the connections in the pool.

carlosms:
So, if I understand your comment, gitbase expires connections after one minute **from creation**, not one minute of inactivity?

ajnavarro:
@carlosms ,no that is what is doing `SetConnMaxLifetime` method

ajnavarro:
I think the correct fix should test the connection before use it. See: https://golang.org/pkg/database/sql/#Conn.PingContext

carlosms:
Push forced a new commit to set max idle connections to 0.

PR #267, opened 2018-10-10
Use go-log lib

### LINKED ISSUE #253: Use go-log
To standardize all our apps, we should use https://github.com/src-d/go-log for logging.
Log options should be managed with https://godoc.org/gopkg.in/src-d/go-cli.v0#LogOptions.

carlosms:
Fix #253.

### DIFF
--- a/cmd/gitbase-web/main.go
+++ b/cmd/gitbase-web/main.go
@@ -2,15 +2,17 @@ package main
 
 import (
 	"database/sql"
+	"encoding/json"
 	"fmt"
 	"net/http"
 
+	"github.com/sirupsen/logrus"
 	"github.com/src-d/gitbase-web/server"
 	"github.com/src-d/gitbase-web/server/handler"
-	"github.com/src-d/gitbase-web/server/service"
 
 	_ "github.com/go-sql-driver/mysql"
 	"gopkg.in/src-d/go-cli.v0"
+	"gopkg.in/src-d/go-log.v1"
 )
 
 // version will be replaced automatically by the CI build.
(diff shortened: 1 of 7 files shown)


cmd/gitbase-web/main.go
```diff
@@ -56,14 +57,35 @@ func (c *ServeCommand) Execute(args []string) error {
 	static := handler.NewStatic("build/public", c.ServerURL, c.SelectLimit, c.FooterHTML)
 
 	// start the router
-	router := server.Router(logger, static, version, db, c.BblfshServerURL)
-	logger.Infof("listening on %s:%d", c.Host, c.Port)
-	err = http.ListenAndServe(fmt.Sprintf("%s:%d", c.Host, c.Port), router)
-	logger.Fatal(err)
+	router := server.Router(logrus.StandardLogger(), static, version, db, c.BblfshServerURL)
```
smola: We should consider `StandardLogger()` for go-log for this kind of use case...

smacker: `StandardLogger` in go-log will return `logrus.Logger`? I'm not sure it's a good idea.

We pass logrus here because the router has request logging middleware that expects logrus interface.

server/handler/query.go
```diff
@@ -128,9 +129,10 @@ func Query(db service.SQLDB) RequestProcessFunc {
 }
 
 func killQuery(db service.SQLDB, query string) {
-	pRows, pErr := db.Query("SHOW FULL PROCESSLIST")
+	const showProcessList = "SHOW FULL PROCESSLIST"
+	pRows, pErr := db.Query(showProcessList)
 	if pErr != nil {
-		// TODO (carlosms) log error when we migrate to go-log
+		log.Errorf(pErr, "failed to execute %q", showProcessList)
```
smacker: it's better to use logger from context. Because currently, it's unclear from logs which request triggered error (in hosted version).

carlosms: Done in d07e77a

cmd/gitbase-web/main.go
carlosms: I have just realized that https://github.com/pressly/lg is deprecated, so we may not want to support this exact use case

smola: @smacker  I thought StandardLogger was meant to return a standard library logger...

Issue #240, opened 2018-09-25
Increase the animation speed for switching tabs


vmarkovtsev:
I switch between tabs with query results and see a nice fade-in/fade-out animation. But after a few queries it becomes annoying - the speed is too low. I suggest increasing the animation speed by at least 2x.

PR #273, opened 2018-10-15
decreasing delay (animations)

chrisgalvan:
From 1000 to 500
Addresses #240


carlosms:
Thanks for your contribution.

Unfortunately this delay is not for the tab switching, this is the delay for the tooltip of the tab titles. Which I fear can become a bit intrusive if we decrease its timeout.

carlosms:
I haven't looked into it yet. There must be a css property we can override.

PR #279, opened 2018-11-05
Update uast-viewer and bblfsh/client-go to work with gitbase 0.18

carlosms:


Gopkg.toml
```diff
@@ -7,7 +7,7 @@
   non-go = true
 
   [[prune.project]]
```
smacker: client-go is pure go project now. We can remove it.

Gopkg.toml
```diff
@@ -39,12 +39,16 @@
   version = "1.2.2"
 
 [[constraint]]
+  name = "gopkg.in/bblfsh/client-go.v3"
+  version = "3.0.0"
```
smacker: 3.1.0

server/handler/uast.go
```diff
@@ -48,37 +48,32 @@ func Parse(bbblfshServerURL string) RequestProcessFunc {
...
-		if resp.Status != protocol.Ok {
-			return nil, serializer.NewHTTPError(http.StatusBadRequest, strings.Join(resp.Errors, "\n"))
+		if err != nil {
+			return nil, serializer.NewHTTPError(http.StatusInternalServerError, err.Error())
 		}
 
-		if resp.UAST != nil && req.Filter != "" {
-			filtered, err := tools.Filter(resp.UAST, req.Filter)
+		if req.Filter != "" {
+			results, err := applyXpath(resp, req.Filter)
```
smacker:
```
resp, err := applyXpath(resp, req.Filter)
```

no need for new variable `results`

server/handler/uast.go
```diff
@@ -106,35 +101,41 @@ func Filter() RequestProcessFunc {
...
-				filtered, err := tools.Filter((*uast.Node)(n), req.Filter)
-				if err != nil {
-					if e, ok := err.(*tools.ErrInvalidArgument); ok {
-						return nil, serializer.NewHTTPError(http.StatusBadRequest, e.Error())
-					}
-					return nil, serializer.NewHTTPError(http.StatusInternalServerError, err.Error())
-				}
-
-				resp.Children = append(resp.Children, filtered...)
+			results, err := applyXpath(reqNodes, req.Filter)
```
smacker: same, no need for `results`

frontend/src/components/UASTViewer.js
```diff
@@ -40,12 +37,27 @@ class UASTViewer extends Component {
...
       .filterUAST(this.props.protobufs, this.state.filter)
       .then(uast => {
-        this.setState({ uast: transformer(uast) });
+        this.setState({ uast: this.transform(uast) });
       })
       .catch(err => this.setState({ uast: null, error: err }))
       .then(() => this.setState({ loading: false }));
   }
 
+  // Applies the uast-viewer object shape transformer, and expands all nodes
```
smacker: before it was expanding only first 2 levels.
The problem with expanding everything - uast is huge. For a big file, it will hang a browser.

Gopkg.toml
carlosms: I forgot to mention I set this to the exact same version as gitbase, to have the same results from the queries and from the parsing we do in our side.
https://github.com/src-d/gitbase/blob/v0.18.0-beta.1/Gopkg.toml#L57

smacker: But your lock file uses 3.1.0:
https://github.com/src-d/gitbase-web/pull/279/commits/540b3ecbe6ca02bb4d004795818b2d67bcc5867a#diff-bd247e83efc3c45ae9e8c47233249f18R316

v3.0.0 doesn't have SyntaxErr.

frontend/src/components/UASTViewer.js
smacker: there is [expandRootIds](https://github.com/bblfsh/uast-viewer/blob/master/src/helpers.js#L105) helper for that. Before it was included in `transformer` but caused some problems so now you need to call it explicitly.

Gopkg.toml
carlosms: :+1: done in 8f25ae5

Gopkg.toml
carlosms: That's right, I had to update to 3.1.0 for the error. It got updated in the lock file but not in this one. Done in 8f25ae5

server/handler/uast.go
carlosms: done in 9bd3a90

frontend/src/components/UASTViewer.js
carlosms: done in a8cacfd

Issue #248, opened 2018-10-01
Update to bblfsh client v3, use UASTv2, add option to choose UAST mode


carlosms:
- [x] gitbase 0.16 made `semantic` UAST the new default. Previously the format was `annotated`. We should also switch to `semantic` for consistency.
- [x] Update to bblfsh/client-go v3, following the gitbase 0.18 beta changes #279
  - [x] Update Dockerfile dependencies #282 
  - [x] Update Travis dependencies #284 
- [x] Use newer uast-viewer release with support for UASTv2 #279
- [x] Add an option to choose the UAST mode (semantic, annotated, native). #283

carlosms:
@ricardobaeta, should we have the same mode selector as in bblfsh web, https://github.com/bblfsh/web/pull/165?

carlosms:
Will leave open until we merge the gitbase-0.18 branch into master.

PR #282, opened 2018-11-09
Update dockerfile dependencies for bblfsh client-go v3

carlosms:
Part of #248.
Updates the dockerfile to remove packages not needed after bblfsh/client-go was updated to v3.


carlosms:
Thanks, I did forget about it completely!
Done in https://github.com/src-d/gitbase-web/pull/284

PR #287, opened 2018-11-22
Service defaults have changed in upstream chart

rporres:
Hence we fix the old defaults to avoid changes in our deployment. The
main thing is to fix service type to node port as it is required by GCE
ingress.

Signed-off-by: Rafael Porres Molina <rafa@sourced.tech>


rporres:
Do not merge until https://github.com/src-d/charts/pull/119 has been merged

PR #298, opened 2018-12-21
Add keyboard shortcuts modal help

### LINKED ISSUE #293: Add a modal help dialog with keyboard shorcuts
This will be a small button that will open a simple modal window with the keyboard shortcuts:
- ctrl+enter to run the query
- ctrl+space to autocomplete.

carlosms:
Fix #293.

@ricardobaeta please comment on icon placement. Also, it would be good if you could provide it as `svg` instead of `png`, like the rest of the icons.

Although it feels to me it is not too subtle for a help icon... One thing we could change: remove this keyboard icon, and have the top-right 'Help' button we already have open a modal with the text:
```
For a reference of gitbase SQL, please read https://docs.sourced.tech/gitbase

Keyboard shortcuts:
[table]
```

image

image

### DIFF
--- a/frontend/src/App.less
+++ b/frontend/src/App.less
@@ -105,7 +105,7 @@ body {
 
 // make modal window "fullscreen"
 // margins are hard-coded in bootstrap, so it's hard-coded here too
-.modal-body {
+.modal-lg .modal-body {
     height: ~'calc(100vh - 150px)';
 
     .uast-viewer {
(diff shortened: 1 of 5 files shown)


ricardobaeta:
Already talked with @carlosms about this on slack and "only the help button, and then inside the modal, the link to the reference and the keyboard shortcut list" seems like an elegant solution to this UI.

PR #300, opened 2019-01-15
Set help link target to blank

carlosms:
The link made in #298 should have had the same `target` as the ones in the sidebar.


marnovo:
@dpordomingo besides opening in a new page? Like some visual cue you mean?

dpordomingo:
visual cue, and also if we should open **always** external links in new tab

PR #305, opened 2019-01-22
Modifies killQuery in order to use the connection id

### LINKED ISSUE #277: Use connection_id to cancel queries
Context in #241, #264, https://github.com/src-d/go-mysql-server/issues/433.

Gitbase 0.18 beta1 added support for `CONNECTION_ID`, so we can use it instead of `SHOW FULL PROCESSLIST`.

se7entyse7en:
Closes #277.

### DIFF
--- a/server/handler/query.go
+++ b/server/handler/query.go
@@ -11,7 +11,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/pressly/lg"
 	"github.com/src-d/gitbase-web/server/serializer"
 	"github.com/src-d/gitbase-web/server/service"
 
@@ -74,8 +73,16 @@ func Query(db service.SQLDB) RequestProcessFunc {
 		c := make(chan error, 1)
 
 		var rows *sql.Rows
+		conn, err := db.Conn(r.Context())
+		defer conn.Close()
+
+		connID, err := getConnID(r, conn)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get connection id: %s", err)
+		}
+
 		go func() {
-			rows, err = db.QueryContext(r.Context(), query)
+			rows, err = conn.QueryContext(r.Context(), query)
 			c <- err
 		}()
 
@@ -88,13 +95,14 @@ func Query(db service.SQLDB) RequestProcessFunc {
 		}
 
 		if r.Context().Err() != nil {
-			killQuery(r, db, query)
+			db.Exec(fmt.Sprintf("KILL %d", connID))
 			return nil, dbError(r.Context().Err())
 		}
 
 		if err != nil {
 			return nil, dbError(err)
 		}
+
 		defer rows.Close()
 
 		columnNames, columnTypes, err := columnsInfo(rows)
@@ -128,46 +136,20 @@ func Query(db service.SQLDB) RequestProcessFunc {
 	}
 }
 
-func killQuery(r *http.Request, db service.SQLDB, query string) {
-	const showProcessList = "SHOW FULL PROCESSLIST"
-	pRows, pErr := db.Query(showProcessList)
-	if pErr != nil {
-		lg.RequestLog(r).WithError(pErr).Errorf("failed to execute %q", showProcessList)
-		return
-	}
-	defer pRows.Close()
+func getConnID(r *http.Request, conn *sql.Conn) (uint32, error) {
+	const connIDQuery = "SELECT CONNECTION_ID()"
+	var connID uint32
 
-	found := false
-	var foundID int
-
-	for pRows.Next() {
-		var id int
-		var info sql.NullString
-		var rb sql.RawBytes
-		// The columns are:
-		// Id, User, Host, db, Command, Time, State, Info
-		// gitbase returns the query on "Info".
-		if err := pRows.Scan(&id, &rb, &rb, &rb, &rb, &rb, &rb, &info); err != nil {
-			lg.RequestLog(r).WithError(err).Errorf("failed to scan the results of %q", showProcessList)
-			return
-		}
-
-		if info.Valid && info.String == query {
-			if found {
-				// Found more than one match for current query, we cannot know which
-				// one is ours. Skip the cancellation
-				lg.RequestLog(r).Errorf("cannot cancel the query, found more than one match in gitbase")
-				return
-			}
-
-			found = true
-			foundID = id
-		}
+	row := conn.QueryRowContext(context.Background(), connIDQuery)
+	if row == nil {
+		return 0, fmt.Errorf("failed to execute %q", connIDQuery)
 	}
 
-	if found {
-		db.Exec(fmt.Sprintf("KILL %d", foundID))
+	if err := row.Scan(&connID); err != nil {
+		return 0, fmt.Errorf("failed to scan the results of %q: %s", connIDQuery, err)
 	}
+
+	return connID, nil
 }
 
 // columnsInfo returns the column names and column types, or error
--- a/server/handler/query_test.go
+++ b/server/handler/query_test.go
@@ -2,6 +2,7 @@ package handler
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
@@ -14,8 +15,9 @@ import (
 
 	"github.com/pressly/lg"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
-	"gopkg.in/DATA-DOG/go-sqlmock.v1"
+	sqlmock "gopkg.in/DATA-DOG/go-sqlmock.v1"
 	"gopkg.in/bblfsh/sdk.v2/uast/nodes"
 )
 
@@ -43,7 +45,7 @@ func (suite *QuerySuite) TestAddLimit() {
 			`, "SELECT * FROM repositories LIMIT 100"},
 		{"  SELECT * FROM repositories  ", "SELECT * FROM repositories LIMIT 100"},
 		{"  SELECT * FROM repositories  ; ", "SELECT * FROM repositories LIMIT 100"},
-		{`  SELECT * FROM repositories  
+		{`  SELECT * FROM repositories
 ; `, "SELECT * FROM repositories LIMIT 100"},
 		{"/* comment */ SELECT * FROM repositories", "SELECT * FROM repositories LIMIT 100"},
 		{"SELECT * FROM repositories /* comment */", "SELECT * FROM repositories LIMIT 100"},
@@ -56,7 +58,7 @@ func (suite *QuerySuite) TestAddLimit() {
 		{"select * from repositories limit 1 ;", "select * from repositories limit 1"},
 		{`select * from repositories limit 1
 ;`, "select * from repositories limit 1"},
-		{`select * from repositories limit 1 
+		{`select * from repositories limit 1
  ; `, "select * from repositories limit 1"},
 		{"select * from repositories limit 900", "select * from repositories LIMIT 100"},
 		{"select * from repositories limit 900;", "select * from repositories LIMIT 100"},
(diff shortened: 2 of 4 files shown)


se7entyse7en:
This PR is not complete yet. I still have to understand how to test it.

se7entyse7en:
Ok, in the end it was not a test issue, but there was actually a bug, that why the expectations weren't met 😅. I was using the same context that was already cancelled to query the connection id, hence the `SELECT CONNECTION_ID()` query was never run.

smacker:
The current code is incorrect.
`SELECT CONNECTION_ID()` will be executed only after the previous query is finished.
The right approach is to take connection id first and only then run the query. With connection id in place, you can kill query using new connection.

se7entyse7en:
@smacker 
> `SELECT CONNECTION_ID()` will be executed only after the previous query is finished.

That was actually made on purpose. The idea is to avoid querying the connection id if it is not necessary and currently is only used when the query is being killed. The connection is then killed using a connection taken from the pool (so a different connection from the one used to query and to extract the connection id).

Am I missing something? Are you perhaps concerned about [this](https://github.com/src-d/gitbase-web/pull/305/files#diff-26acfe363b40e46e45577cb3ad576f01R138) where I use a newly created context as the one corresponding to the request has been cancelled?

If I do as you say there's no actually need to use a different context at the cost of always querying the connection id.

smacker:
Hm. Maybe I misunderstood something but I suppose it works like this now:

http request -> create connection -> execute query -> http cancel -> query is still running in connection -> query connection_id waits for running query to finish (because it's executed in the same connection to get correct id).

Please correct me if I'm wrong. (like if context cancelation actually breaks the connection to db but then I'm not sure that you get a correct id...)

se7entyse7en:
> query connection_id waits for running query to finish (because it's executed in the same connection to get correct id)

You got me here! I missed the fact the the `SELECT CONNECTION_ID()` is being enqueued given that it uses the same connection. So yes, the only viable solution is to query the connection id first. 👍 

Thanks @smacker

server/handler/query.go
```diff
@@ -74,8 +74,17 @@ func Query(db service.SQLDB) RequestProcessFunc {
 		c := make(chan error, 1)
 
 		var rows *sql.Rows
+		conn, err := db.Conn(r.Context())
+		defer conn.Close()
```
carlosms: If we close after each query, does it negate the `GITBASEPG_CONN_MAX_LIFETIME` config? I mean that the connection pool will not contain idle connections ever, right?

server/handler/query.go
```diff
@@ -128,46 +137,20 @@ func Query(db service.SQLDB) RequestProcessFunc {
...
-				// one is ours. Skip the cancellation
-				lg.RequestLog(r).Errorf("cannot cancel the query, found more than one match in gitbase")
-				return
-			}
-
-			found = true
-			foundID = id
+	if row := conn.QueryRowContext(context.Background(), connIDQuery); row != nil {
+		if err := row.Scan(&connID); err != nil {
+			lg.RequestLog(r).WithError(err).Errorf("failed to scan the results of %q", connIDQuery)
```
carlosms: since `Query()` is already going to log the err, and for consistency with the other error returned in this method, I would change this to
```go
return 0, fmt.Errorf("failed to scan the results of %q": %s", connIDQuery, err)
```

server/handler/query.go
```diff
@@ -84,17 +93,17 @@ func Query(db service.SQLDB) RequestProcessFunc {
 		// case. We check if the context was cancelled with Err() instead of Done()
 		select {
 		case <-r.Context().Done():
+			if r.Context().Err() != nil {
```
carlosms: Is it safe to move this here from the previous place, after the `select{}`?

As the comment above says, if the context is cancelled this select may get into the `case err` branch. In that case, the `KILL` will not be executed.

server/handler/query.go
se7entyse7en: > If we close after each query, does it negate the `GITBASEPG_CONN_MAX_LIFETIME` config?

It shouldn't. By reading the docs and the article I linked [here](https://github.com/src-d/gitbase-web/pull/306#discussion_r250685778), the connection must be closed in order to make it available again in the pool. From the doc:

> Every Conn must be returned to the database pool after use by calling Conn.Close.

I *guess* that without closing it, it just becomes some sort of opened zombie connection. This leak could lead the pool to consider all the connections as "in-use" and reach the limit number of opened connections. By considering them "in-use" I also think that the connection won't be considered expired and won't honour `GITBASEPG_CONN_MAX_LIFETIME`.

Anyway I'm not sure about all this, but maybe I can try doing a simple test where I open a connection without closing it and see the [stats of the pool](https://golang.org/pkg/database/sql/#DBStats).

carlosms: Thanks, I was assuming a different meaning for `Conn.Close()` :+1:

se7entyse7en: Okay, just to be sure I tested id 😛. Here's the code that I ran:

```package main

import (
	"database/sql"
	"fmt"
	"net/http"

	_ "github.com/mattn/go-sqlite3"
)

func main() {
	db, err := sql.Open("sqlite3", "./foo.db")
	db.SetMaxOpenConns(3)

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		fmt.Println("===START===")
		fmt.Fprintf(w, "Welcome to my website!")
		if err != nil {
			fmt.Println(err)
			return
		}

		fmt.Printf("Stats: %v\n", db.Stats())
		fmt.Println("Opening connection...")

		conn, err := db.Conn(r.Context())
		if err != nil {
			fmt.Println(err)
			return
		}

		fmt.Printf("Stats: %v\n", db.Stats())
		fmt.Println("Getting connection from pool")

		fmt.Println("Connection opened")
		fmt.Println("Pinging connection...")
		conn.PingContext(r.Context())
		fmt.Println("Avoided closing connection")
		fmt.Println("===END===")
	})

	http.ListenAndServe(":80", nil)
}
```

And here's the screen of the ouput:

screenshot 2019-01-28 at 19 15 08

After the 3rd request, the server is not able to open new connections, and the request hangs indefinitely. I tried to add the `defer conn.Close()` and the stats shows the swtich between idle and in use at each request.

se7entyse7en: Actually in our case I don't think we have a limit on the number of connections, but it's clearly a leak 👍

server/handler/query.go
se7entyse7en: I explored what's happening here and I also tried to simulate it [here](https://play.golang.org/p/aj8afgoaiCP).

There are 3 possible outcomes:
1. The query runs and returns no error,
2. The query runs and returns an error,
3. The query runs and is cancelled by the user.

In the first case the `select` unblocks as it receives `nil` in channel `c`.
In the second case the `select` unblocks as it receives a non-`nil` value in channel `c`.
In the third case the `select` unblocks as the cancelled context becomes `Done`, and in this case `r.Context().Err() != nil`. Additionally from the snippet that I wrote it the `err` channel is even never written in this case (unless the outcome of the writing is not deterministic, could that be the case?).

server/handler/query.go
```diff
@@ -128,46 +137,20 @@ func Query(db service.SQLDB) RequestProcessFunc {
...
-			}
-
-			found = true
-			foundID = id
+	if row := conn.QueryRowContext(context.Background(), connIDQuery); row != nil {
+		if err := row.Scan(&connID); err != nil {
+			lg.RequestLog(r).WithError(err).Errorf("failed to scan the results of %q", connIDQuery)
+			return 0, err
 		}
+	} else {
```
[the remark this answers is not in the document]
se7entyse7en: Yup, I always forget that 😕. Thanks!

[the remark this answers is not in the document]
se7entyse7en: Addressed [here](https://github.com/src-d/gitbase-web/pull/305/commits/c978a994ee8f289697ef03f292abecde96d0e09a)

server/handler/query.go
se7entyse7en: Addressed [here](https://github.com/src-d/gitbase-web/pull/305/commits/1ed3bc62e2e9474f72fedae0e40277fccb1f23ae)

server/handler/query.go
se7entyse7en: Addressed [here](https://github.com/src-d/gitbase-web/pull/305/commits/1ed3bc62e2e9474f72fedae0e40277fccb1f23ae)

server/handler/query.go
carlosms: Thank you for the thorough follow-up. In our case we should have the [default number of idle connections, 2](https://golang.org/pkg/database/sql/#DB.SetMaxIdleConns).

server/handler/query.go
carlosms: For scenario 3, what can happen is that the `conn.QueryContext` will stop when the context is cancelled and return an error. Then in this select, you will enter any of the 2 possible `case`, and it's not deterministic. From the [go docs](https://tour.golang.org/concurrency/5): 

> A select blocks until one of its cases can run, then it executes that case. It chooses one at random if multiple are ready.

You can test this in your branch, run this:
```
watch -n 0.1 -e "GOCACHE=off go test -v github.com/src-d/gitbase-web/server/handler -run TestQuerySuite/TestQueryAbort"
```

And sooner or later you will get a failure:
```
=== RUN   TestQuerySuite
=== RUN   TestQuerySuite/TestQueryAbort
--- FAIL: TestQuerySuite (0.20s)
    --- FAIL: TestQuerySuite/TestQueryAbort (0.20s)
        require.go:157:
                Error Trace:    query_test.go:244
                                                        server.go:1964
                                                        request_logger.go:39
                                                        server.go:1964
                                                        query_test.go:249
                                                        asm_amd64.s:1333
                Error:          Not equal:
                                expected: *errors.errorString(&errors.errorString{s:"context canceled"})
                                actual  : serializer.httpError(serializer.httpError{Status:400, Title:"canceling query due to user request", Details:"", MySQLCode:0x0})
                Test:           TestQuerySuite/TestQueryAbort
        common_unit_test.go:45:
                Error Trace:    common_unit_test.go:45
                                                        suite.go:98
                                                        suite.go:103
                Error:          there were unfulfilled expectations:
                Test:           TestQuerySuite/TestQueryAbort
                Messages:       there is a remaining expectation which was not matched: ExpectedExec => expecting Exec or ExecContext which:
                                  - matches sql: 'KILL 1288'
                                  - is without arguments
FAIL
FAIL    github.com/src-d/gitbase-web/server/handler     0.229s
```

Note: the `Not equal` check is correct, but error message is wrong. It looks like we have the arguments for `require.Equal(context.Canceled, err)` reversed in the test, it should be `require.Equal(err, context.Canceled).

se7entyse7en: Good think you noticed! I haven't thought that the context could be cancelled just **after** the channel has been written, but **before** entering the select! I'm gonna fix it and thanks for explanation @carlosms!

se7entyse7en: Addressed [here](https://github.com/src-d/gitbase-web/pull/305/commits/7e62f4436c93423abf5193a9a5611b4b54cf5bc3).

se7entyse7en:
Gonna fix the DCO in final squash and rebase.

server/handler/query.go
```diff
@@ -74,8 +74,17 @@ func Query(db service.SQLDB) RequestProcessFunc {
 		c := make(chan error, 1)
 
 		var rows *sql.Rows
+		conn, err := db.Conn(r.Context())
+		defer conn.Close()
+
+		connID, err := getConnID(r, conn)
+		if err != nil {
+			lg.RequestLog(r).WithError(err).Error("failed to get connection id")
```
carlosms: I think we already log the returned errors in `render.go`, so I would do:
```go
return nil, fmt.Errorf("failed to get connection id: %s", err)
```

server/handler/query.go
```diff
@@ -128,46 +138,21 @@ func Query(db service.SQLDB) RequestProcessFunc {
...
-		}
+	row := conn.QueryRowContext(context.Background(), connIDQuery)
+	if row == nil {
+		return 0, fmt.Errorf("failed to execute %q", connIDQuery)
 	}
 
-	if found {
-		db.Exec(fmt.Sprintf("KILL %d", foundID))
+	if err := row.Scan(&connID); err != nil {
+		lg.RequestLog(r).WithError(err).Errorf("failed to scan the results of %q: %s", connIDQuery, err)
```
carlosms: Same comment, let's return the err and log only in one place.

se7entyse7en:
@carlosms addressed both [here](https://github.com/src-d/gitbase-web/pull/305/commits/8f6ae7bb3ef64b4a922bc85bb8def879d26374ff).

PR #306, opened 2019-01-23
Adds possibility to configure conn max lifetime

### LINKED ISSUE #262: Implement db idle connections with a default SetConnMaxLifetime
PR #261 was needed because after an idle time of >60s the next query was always failing with `Query Failed - invalid connection`.

The quick fix was to set the max number of idle connections to 0, but his negates the purpose of having a connection pool. We should have a reasonable default for [`SetConnMaxLifetime`](https://golang.org/pkg/database/sql/#DB.SetConnMaxLifetime). If the connection expiration time can be configured in gitbase, maybe we should also have an option in gitbase-web.

se7entyse7en:
Closes #262.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ Any of the previous execution methods accept configuration through the following
 | `GITBASEPG_PORT` | `--port` | `8080` | Port to bind the HTTP server |
 | `GITBASEPG_SERVER_URL` | `--server` | | URL used to access the application in the form `HOSTNAME[:PORT]`. Leave it unset to allow connections from any proxy or public address |
 | `GITBASEPG_DB_CONNECTION` | `--db` | `root@tcp(localhost:3306)/none?maxAllowedPacket=4194304` | gitbase connection string. Use the DSN (Data Source Name) format described in the [Go MySQL Driver docs](https://github.com/go-sql-driver/mysql#dsn-data-source-name). |
+| `GITBASEPG_CONN_MAX_LIFETIME` | `--conn-max-lifetime` | `30` | Maximum amount of time a SQL connection may be reused, in seconds. Make sure this value is lower than the timeout configured in the gitbase server, set with [`GITBASE_CONNECTION_TIMEOUT`](https://docs.sourced.tech/gitbase/using-gitbase/configuration#environment-variables) |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -76,6 +76,7 @@ Any of the previous execution methods accept configuration through the following
 | `GITBASEPG_PORT` | `--port` | `8080` | Port to bind the HTTP server |
 | `GITBASEPG_SERVER_URL` | `--server` | | URL used to access the application in the form `HOSTNAME[:PORT]`. Leave it unset to allow connections from any proxy or public address |
 | `GITBASEPG_DB_CONNECTION` | `--db` | `root@tcp(localhost:3306)/none?maxAllowedPacket=4194304` | gitbase connection string. Use the DSN (Data Source Name) format described in the [Go MySQL Driver docs](https://github.com/go-sql-driver/mysql#dsn-data-source-name). |
+| `GITBASEPG_CONN_MAX_LIFETIME` | `--conn-max-lifetime` | `30` | Connections max life time since their creation in seconds |
```
carlosms: Let's add a mention to the timeout config in gitbase. Something like
```suggestion
| `GITBASEPG_CONN_MAX_LIFETIME` | `--conn-max-lifetime` | `30` | Maximum amount of time a SQL connection may be reused, in seconds. Make sure this value is lower than the timeout configured in the gitbase server, set with [`GITBASE_CONNECTION_TIMEOUT`](https://docs.sourced.tech/gitbase/using-gitbase/configuration#environment-variables) |
```

dpordomingo: Is it reasonable for this web UI to wait for more than a couple of minutes?
On the other hand, if that's the use case of this playground, let's do it.
@smacker do you remember an average of your slow queries?

se7entyse7en: @smacker that's a nice point, I forgot to tell that the chosen default was not inspired by any particular thought. Anyway the fact that some queries are taking longer shouldn't affect this value as non-idle connection can still be in-use after the timeout. It just cannot be started after the timeout. At least this is what I read from [here](https://www.alexedwards.net/blog/configuring-sqldb).

> A connection can still be in use more than <timeout value> after being created — it just cannot start to be reused after that time.

smacker: oh, nice! Thank you for the clarification! 👍

Issue #311, opened 2019-02-14
Update dependencies and enable dependabot


carlosms:

carlosms:
Done, dependabot is not configured to bump go & JS dependencies weekly.

PR #312, opened 2019-02-14
Update go dependencies

carlosms:
Part of #311.
I didn't update https://github.com/rs/cors because they have newer tags, but the releases page still marks 1.3.0 as the latest release.


smacker:
Could you please check what new enry returns for C#? And if it matches what we expect in language selector.

carlosms:
Does not change, it's `C#` in both versions.
I created an issue for your other question #314.

PR #313, opened 2019-02-14
Modifies Makefile in order to include bin for OSX and Windows

### LINKED ISSUE #301: Build and release binaries for macOS and windows
Older versions of libuast prevented us from building other platform that linux. Now that we use bblfsh/client-go v3 (#279) we should be able to build and release the binaries for macOS and windows.

se7entyse7en:
Closes #301.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ DEPENDENCIES := \
 	golang.org/x/lint/golint
 DEPENDENCIES_DIRECTORY := ./vendor
 
-PKG_OS = linux
+PKG_OS = linux darwin windows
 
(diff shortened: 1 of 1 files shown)


carlosms:
Did you test if the binaries built on linux work for macOS and windows?

smacker:
why is it necessary to test without cgo?

carlosms:
why not? It _should_ work, but it would be nice to know that it works as expected.

smacker:
idk. Go guarantees it will work as long as we don't use cgo and don't rely on any external dependencies (like docker for engine) or don't use anything os specific (like file paths in lookout)

se7entyse7en:
@carlosms I've tested the OSX binary, but I haven't the windows one.

PR #320, opened 2019-02-20
Update gitbase and bblfsh

carlosms:


.helm-defaults.yml
```diff
@@ -36,7 +36,7 @@ bblfshdServer:
         tag: v2.5.0
       csharp:
         repository: bblfsh/csharp-driver
-        tag: v1.2.0
+        tag: v1.3.0
```
smacker: there is v1.4.0 already which is considered real 1.0.

PR #348, opened 2019-03-05
Update docker-compose.yml and README

mcarmonaa:
Related to the empathy-session.

- README: fix `docker-compose up` command description.
- docker-compose.yml: remove `GITBASE_UNSTABLE_SQUASH_ENABLE: "true"` line. This is the default behavior since gitbase v0.14.0. Make the directory `./repos` the default one if there's no one set.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,6 @@ services:
     image: "srcd/gitbase:v0.18.0"
     environment:
       BBLFSH_ENDPOINT: bblfsh:9432
-      GITBASE_UNSTABLE_SQUASH_ENABLE: "true"
     volumes:
       - ${GITBASEPG_REPOS_FOLDER}:/opt/repos
   bblfsh:
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ $ git clone git@github.com:bblfsh/bblfshd.git
 $ git clone git@github.com:src-d/gitbase-web.git
 ```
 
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -14,9 +14,8 @@ services:
     image: "srcd/gitbase:v0.18.0"
     environment:
       BBLFSH_ENDPOINT: bblfsh:9432
-      GITBASE_UNSTABLE_SQUASH_ENABLE: "true"
     volumes:
-      - ${GITBASEPG_REPOS_FOLDER}:/opt/repos
+      - ${GITBASEPG_REPOS_FOLDER-./repos}:/opt/repos
```
carlosms: I'm not sure about this change. The docs mention `$HOME/repos`, not `./repos`.
Also, not having a default value forces the first-time user to pay attention and read the docs. Preparing a directory with a few repos is mandatory, and starting the containers with no errors but also no data can be more confusing.

WDYT @dpordomingo?

dpordomingo: I agree with @carlosms 
It's a good idea to avoid starting the playground without repositories in the DB, or at least without forcing the user to explicitly define a place.

about using `./repos` or `$HOME/repos`, I'm ok with both, but it should be consistent with docs.

mcarmonaa:
I've reverted the changes about the `GITBASEPG_REPOS_FOLDER`

PR #358, opened 2019-03-06
Kill queries also after the rows start streaming

### LINKED ISSUE #357: Cannot cancel queries that are already streaming results
We are cancelling queries correctly if gitbase still has not responded with the results. But when gitbase starts streaming rows, our query cancellation leaves the query running on gitbase.

For example, `select * from files order by file_path` can be cancelled because gitbase does not return until it finishes the ordering. But `select * from files` starts to collect the rows quickly, and if you cancel after a time the query continues in gitbase. The problem is [this `select`](https://github.com/src-d/gitbase-web/blob/master/server/handler/query.go#L92).

carlosms:
Fix #357.

See issue for details on the bug. This PR basically makes bigger the "wrapper logic" of killing the connection on context cancellation from (`QueryContext`) to (`QueryContext` + rows processing).

### DIFF
--- a/server/handler/query.go
+++ b/server/handler/query.go
@@ -53,36 +53,37 @@ func genericVals(colTypes []string) []interface{} {
 // the rows as JSON
 func Query(db service.SQLDB) RequestProcessFunc {
 	return func(r *http.Request) (*serializer.Response, error) {
-		var queryRequest queryRequest
+		var queryReq queryRequest
 		body, err := ioutil.ReadAll(r.Body)
 		if err != nil {
 			return nil, err
 		}
 
-		err = json.Unmarshal(body, &queryRequest)
-		if err != nil || queryRequest.Query == "" {
+		err = json.Unmarshal(body, &queryReq)
+		if err != nil || queryReq.Query == "" {
 			return nil, serializer.NewHTTPError(http.StatusBadRequest,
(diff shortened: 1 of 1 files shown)


server/handler/query.go
```diff
@@ -103,40 +104,53 @@ func Query(db service.SQLDB) RequestProcessFunc {
...
-		defer rows.Close()
+		return resp, nil
+	}
+}
 
-		columnNames, columnTypes, err := columnsInfo(rows)
-		if err != nil {
-			return nil, err
-		}
+func queryContext(ctx context.Context, conn *sql.Conn, queryRequest queryRequest) (*serializer.Response, error) {
```
dpordomingo: is it really neede to shadow `queryRequest` type?

carlosms: sorry, I don't follow what you mean here. Can you elaborate?

smacker: the name of the parameter is `queryRequest` and the type of this parameter is `queryRequest`.
because of it you don't have access to the type anymore inside the function and it is a little bit confusing that `queryRequest` outside of this function is a type but inside an instance of this type.

carlosms: you are right, I just copied the code we had before and didn't realize that. Thanks.
I've renamed the var, and because it's a small change I already squashed it to make the branch ready to merge.

PR #395, opened 2019-04-03
Update bblfshd to v2.12.0

carlosms:


.helm-defaults.yml
```diff
@@ -5,38 +5,38 @@ service:
...
-    tag: v2.11.8
+    tag: v2.12.0
     pullPolicy: IfNotPresent
   drivers:
     install: true
     languages:
       python:
         repository: bblfsh/python-driver
-        tag: v2.7.0
+        tag: v2.6.0
```
se7entyse7en: Why this is downgraded?

carlosms: oops. Fixed and squashed, thanks!

PR #397, opened 2019-04-05
[Update create-react-app to 2.x

carlosms:
This PR replaces #361.

I followed the migration guide from https://github.com/facebook/create-react-app/releases/tag/v2.0.3.

react-app-rewired does not fully support Create React App 2.0. We used react-app-rewired for SVG loading, and less support. 

Create React App 2.x already supports SVGs.
For less I installed https://github.com/sharegate/craco and https://github.com/FormAPI/craco-less, which seems to be a popular alternative to react-app-rewired.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -4,6 +4,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@craco/craco": "^4.1.0",
     "abortcontroller-polyfill": "^1.3.0",
     "bootstrap": "3",
     "codemirror": "^5.45.0",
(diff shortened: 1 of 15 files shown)


frontend/package.json
```diff
@@ -45,30 +44,10 @@
...
-      "target": "http://localhost:8080"
-    },
-    "/filter": {
-      "target": "http://localhost:8080"
-    },
-    "/version": {
-      "target": "http://localhost:8080"
-    }
-  }
+  "browserslist": [
```
smacker: why do we need to define it here? I don't see anything about browserslist in the migration guide. Does `create-react-app` have some any defaults?

frontend/src/setupProxy.js
```diff
@@ -0,0 +1,12 @@
+const proxy = require('http-proxy-middleware');
+
+module.exports = function(app) {
+  app.use(proxy('/query', { target: 'http://localhost:8080' }));
```
smacker: we are not limited by json anymore. Let's make it a bit prettier:
```js
['/query',  '/export', ...].forEach(path => app.use(proxy(path, { target: 'http://localhost:8080' })))
```

frontend/package.json
carlosms: The first time you run `yarn start` it complains that `browserslist` is not defined, and offers to fill it with the defaults you see here. I just left the defaults.

frontend/src/setupProxy.js
carlosms: :+1: done in 8a440da.

PR #403, opened 2019-04-11
Migrates from client-go to go-client

### LINKED ISSUE #401: Migrate from bblfsh/client-go to bblfsh/go-client
related to src-d/engine/issues/418
related to bblfsh/web/issues/208

https://github.com/bblfsh/client-go is now deprecated, we need to start using https://github.com/bblfsh/go-client.

se7entyse7en:
Closes #401.


se7entyse7en:
Please take good look at this as I'm not very knowledgable in go deps handling 😞. As in this case I don't know if I've done the right thing, as the import should also have the `/v4` suffix as written in the README.

se7entyse7en:
> /v4 is a go mod suffix. dep doesn't know anything about it. Go 1.11/12 also wouldn't be able to import such path without GO111MODULE=on.

Thanks for the explanation!

PR #410, opened 2019-04-16
Ran dep ensure -update

se7entyse7en:
Kind of related to [this](https://github.com/src-d/gitbase-web/pull/403).

I hadn't run `dep ensure -update` in that PR.


carlosms:
I wonder why there are updates if we are using dependabot to keep up to date.

se7entyse7en:
mhmmm... I didn't think about dependabot, but yup I'd expect the changes to involve only `go-client`, but it seems that a lot of other things have changed.

carlosms:
I'm not sure if upgrading this many things at once is risky. Did you go over the release notes?

se7entyse7en:
Sorry, maybe I still didn't understand `dep` too. But AFAIU `dep ensure` updates `vendor` from `Gopkg.toml` and `Gopkg.lock`, but to update `Gopkg.lock` we have to run `dep ensure -update`. So didn't we need to always run `dep ensure -update` every time we add a dependency?

se7entyse7en:
Now one thing that I don't understand is why [here](https://github.com/src-d/gitbase-web/pull/403) the `vendor` doesn't have any updates.

se7entyse7en:
> Now one thing that I don't understand is why here the vendor doesn't have any updates.

Oh ok I realized that `go-client` was already present in `vendor`.

smacker:
```
dep ensure -update
```
updates all the dependencies

```
dep ensure -update package
```
updates only one package.

```
$ dep ensure --help
...
  -update       update the named dependencies (or all, if none are named) in Gopkg.lock to the latest allowed by Gopkg.toml (default: false)
```

se7entyse7en:
Okay I also realized that this

> Now one thing that I don't understand is why here the vendor doesn't have any updates.

wasn't true, it's clear that the package in `vendor` has been renamed. Okay gonna close this! 👍 Sorry for the mess.

PR #412, opened 2019-04-18
Adds TestUASTGetLanguagesSuite/TestSameEnryLanguage

### LINKED ISSUE #314: Write test to check the language names returned by bblfsh and enry match
As requested by @smacker [here](https://github.com/src-d/gitbase-web/pull/312#issuecomment-463982604)
> Could you please check what new enry returns for C#? And if it matches what we expect in language selector.

The goal is to have a test to check if the output of `/get-languages` uses the same names as `/detect-lang` (i.e. `csharp` vs `C#).

se7entyse7en:
Closes #314.

This test ensures that the language label returned by `enry` is the same
used by `bblfsh`.

### DIFF
--- a/server/handler/uast_integration_test.go
+++ b/server/handler/uast_integration_test.go
@@ -2,18 +2,155 @@ package handler_test
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
 
+	"github.com/go-chi/chi"
 	"github.com/pressly/lg"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/suite"
+	"gopkg.in/src-d/enry.v1"
 
 	"github.com/src-d/gitbase-web/server/handler"
 	"github.com/src-d/gitbase-web/server/serializer"
+	"github.com/src-d/gitbase-web/server/service"
 )
+
+type UASTGetLanguagesSuite struct {
+	suite.Suite
+	handler http.Handler
+}
+
+func TestUASTGetLanguagesSuite(t *testing.T) {
+	if !isIntegration() {
+		t.Skip("use the env var GITBASEPG_INTEGRATION_TESTS=true to run this test")
+	}
+
+	q := new(UASTGetLanguagesSuite)
+	r := chi.NewRouter()
+	r.Use(lg.RequestLogger(logrus.New()))
+	r.Post("/detect-lang", handler.APIHandlerFunc(handler.DetectLanguage()))
+	r.Get("/get-languages", handler.APIHandlerFunc(handler.GetLanguages(bblfshServerURL())))
+
+	q.handler = r
+
+	suite.Run(t, q)
+}
+
+func UnmarshalGetLanguagesResponse(b []byte) []service.Language {
+	var resBody struct {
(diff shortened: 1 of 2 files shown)


server/handler/uast_integration_test.go
```diff
@@ -2,19 +2,156 @@ package handler_test
...
+			res = httptest.NewRecorder()
+			suite.handler.ServeHTTP(res, req)
+
+			require.Equal(http.StatusOK, res.Code, res.Body.String())
+
+			detectedLang, detectedLangType := handler.UnmarshalDetectLangResponse(res.Body.Bytes())
+
+			if langName == "Bash" {
+				require.NotEqual(langName, detectedLang)
+				t.Skip("TEST FAILURE IS A KNOWN ISSUE")
```
se7entyse7en: Is this correct? Isn't this due to the aliasing problem that we also have on engine? But shouldn't it be true also for `C#`?

server/handler/uast_integration_test.go
```diff
@@ -2,19 +2,156 @@ package handler_test
...
+		suite.T().Run(langName, func(t *testing.T) {
+			content, filename := suite.getContentAndFilename(langName)
+			jsonRequest := fmt.Sprintf(`{ "content": "%s", "filename": "%s" }`,
+				escapeForJSON(content), filename)
+			req, _ := http.NewRequest("POST", "/detect-lang", strings.NewReader(jsonRequest))
+
+			res = httptest.NewRecorder()
+			suite.handler.ServeHTTP(res, req)
+
+			require.Equal(http.StatusOK, res.Code, res.Body.String())
```
carlosms: this `require` is not using the current `t *testing.T`

se7entyse7en: oops, fixed [here](https://github.com/src-d/gitbase-web/pull/412/commits/30df0ebe04dda15f39826a7098d910665a53cfd2).

server/handler/uast_integration_test.go
```diff
@@ -2,19 +2,157 @@ package handler_test
...
+	}
+
+	for _, lang := range UnmarshalGetLanguagesResponse(res.Body.Bytes()) {
+		langName := lang.Name
+		suite.T().Run(langName, func(t *testing.T) {
+			require := require.New(t)
+
+			content, filename := suite.getContentAndFilename(langName)
+			jsonRequest := fmt.Sprintf(`{ "content": "%s", "filename": "%s" }`,
+				escapeForJSON(content), filename)
```
carlosms: Instead of this manual escape you could do
```go
			j := struct {
				Filename string
				Content  string
			}{
				Filename: filename,
				Content:  content,
			}

			b, _ := json.Marshal(j)
			req, _ := http.NewRequest("POST", "/detect-lang", bytes.NewReader(b))
```

or export `detectLangRequest` type and
```go
			j := handler.DetectLangRequest{
				Filename: filename,
				Content:  content,
			}

			b, _ := json.Marshal(j)
			req, _ := http.NewRequest("POST", "/detect-lang", bytes.NewReader(b))
```

se7entyse7en: thanks for the tip! Done [here](https://github.com/src-d/gitbase-web/pull/412/commits/b1a304cf5b4548b891612f47031ad8748f825ad9).

PR #439, opened 2019-10-02
Remove Drone

meyskens:
This removes the Drone integration as we're decommissioning Drone


rporres:
cc @src-d/applications

dpordomingo:
Are we still supporting this project? Didn't it stale when we started working in source{d}?

rporres:
That's why we're not migrating it to Travis, @dpordomingo

dpordomingo:
If we merge, http://gitbase-playground-staging.srcd.run will become stale and we won't be able to deploy a new version, what lgtm.

I wonder if we could also turn the service itself off, or we might confirm it in advance with @src-d/product

